PR #52, opened 2016-02-25
Fix benchmark PrintResults

buybackoff:
Couldn't believe the numbers at the first glance, but sadly there was a typo.


ColinSullivan1:
Yep, certainly a typo there - thank you for the contribution!

PR #68, opened 2016-05-23
trigger a reconnect in case of error sent by the server

stephenjannin:
Example : Authorization timeout etc...
Never leave a connection in status CLOSED

### DIFF
--- a/NATS/Conn.cs
+++ b/NATS/Conn.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2015 Apcera Inc. All rights reserved.
+// Copyright 2015 Apcera Inc. All rights reserved.
 
 using System;
 using System.Collections.Generic;
@@ -1753,17 +1753,7 @@ namespace NATS.Client
             else
             {
                 ex = new NATSException(s);
-                lock (mu)
-                {
-                    lastEx = ex;
-
-                    if (status != ConnState.CONNECTING)
-                    {
-                        invokeDelegates = true;
-                    }
-                }
-
-                close(ConnState.CLOSED, invokeDelegates);
+				processOpError(ex);
             }
         }
 


ColinSullivan1:
Thanks @stephenjannin, This should fix it a potential sticky closed status issue, but I'm looking through to see if there is a different place, as to be more congruent with the go client.  Would you have a unit test?

ColinSullivan1:
I've added a few tests to verify the client can reconnect after receiving an authorization timeout, both with standard connections and TLS connections,  [here](https://github.com/nats-io/csnats/blob/master/NATSUnitTests/UnitTestAuth.cs#L128) and [here](https://github.com/nats-io/csnats/blob/master/NATSUnitTests/UnitTestTLS.cs#L272).  They succeed for me.  Could you verify these work for you with the current source in master?

Looking at the current source, an authorization error returned by the server would result in an exception being thrown by processConnInit, and handled where the connection status is reset to RECONNECTING [here](https://github.com/nats-io/csnats/blob/master/NATS/Conn.cs#L1303).  On receiving other errors from the server, it's also set [here](https://github.com/nats-io/csnats/blob/master/NATS/Conn.cs#L1159).  Is it possible your client with the issue was using older source, before the recently added defensive code was added to reconnect?

stephenjannin:
My services are based on commit e3e2c850174e8a88395aa18aa3dfba85daacc9bc, so it's already including pull request #63 . I Cherry-picked your pull request #70, but it's not reproducing the issue as we don't break into Connection.processErr(MemoryStream errorStream).

You can try this code in TestTLS.TestTlsReconnectAuthTimeoutFailure(), juste before s1.ShutDown(). The test will fail

```
            MethodInfo processErr = typeof(Connection).GetMethod("processErr", BindingFlags.NonPublic | BindingFlags.Instance);
            using (var ms = new MemoryStream())
            {
                var bytes = System.Text.Encoding.UTF8.GetBytes("fake server error");
                ms.Write(bytes,0,bytes.Length);
                ms.Position = 0;
                processErr.Invoke(c, new object[] { ms });
            }
```

ColinSullivan1:
@stephenjannin, I see...  The timing your are experiencing, likely under a stressed system, results in the client processing the authorization timeout before the connection disconnect.  Receiving an authorization timeout should trigger a reconnect, but does not today as we rely on the connection being closed by the server.  I ran 1000 iterations of TestTlsReconnectAuthTimeoutFailure() and did not encounter this, so this must be interesting timing between the server and client.  I'll look into this further.

stephenjannin:
I think you're right.
I have a 3 nodes gnatsd cluster, with >250 clients per node. SSL + authorization are activated.
When performing a rolling upgrade of the gnatsd cluster, a node will have to handle > 250 clients connections before the configured timeout (5 seconds). This can take some time !

ColinSullivan1:
@stephenjannin, This was great test code - thank you!  Your fix would work for this situation, but we'd rather allow only certain errors from the server to start the client reconnection process, to favor protecting the server over clients.  Could you let me know if the changes in #72 resolves your problem?

stephenjannin:
Thank you for the fix.
Everything is upgraded on our side. I'll do more nats cluster crash test tomorrow.

stephenjannin:
Bad news, the fix is not working. I have 13/84 services that didn't survived the cluster restart : 

20160526 09:50:51,029 [INFO ] NatsReconnected : CONNECTED LastError=  
20160526 09:51:34,624 [INFO ] NatsReconnected : CONNECTED LastError=  
20160526 09:52:00,002 [WARN ] GetConnectionWarning : CLOSED LastError=NATS.Client.NATSException: Authorization Timeout'  
20160526 09:52:00,006 [INFO ] ConnectionStats : CLOSED inBytes=19782 outBytes=455302 inMsgs=30 outMsgs=686 reconnects=2 
20160526 09:52:00,016 [ERROR] SendHeartBeat(): Connection is closed.
   at NATS.Client.Connection.publish(String subject, String reply, Byte[] data)

stephenjannin:
ok, I got it. See last Exception of NATS.Client.Connection, there is a ' at the end of exception message !!!!

I will patch the driver in Parser.cs with  case '\'': : 

```
                case MINUS_ERR_ARG:
                    switch (b)
                    {
                        case '\'':
                        case '\r':
                            break;
                        case '\n':
                            conn.processErr(argBufStream);
                            argBufStream.Position = 0;
                            state = OP_START;
                            break;
                        default:
                            argBufStream.WriteByte((byte)b);
                            break;
                    }
                    break;
```

0:000> !mdt 0000000380070220
0000000380070220 (NATS.Client.NATSException)
    _className:00000000800ccf50 (System.String) Length=25, String="NATS.Client.NATSException"
    _exceptionMethod:NULL (System.Reflection.MethodBase)
    _exceptionMethodString:NULL (System.String)
    _message:00000003800701d8 (System.String) Length=22, String="Authorization Timeout'"
    _data:NULL (System.Collections.IDictionary)
    _innerException:NULL (System.Exception)
    _helpURL:NULL (System.String)
    _stackTrace:NULL (System.Object)
    _watsonBuckets:NULL (System.Object)
    _stackTraceString:NULL (System.String)
    _remoteStackTraceString:NULL (System.String)
    _remoteStackIndex:0x0 (System.Int32)
    _dynamicMethods:NULL (System.Object)
    _HResult:0x80131500 (System.Int32)
    _source:NULL (System.String)
    _xptrs:0000000000000000 (System.IntPtr)
    _xcode:0xe0434352 (System.Int32)
    _ipForWatsonBuckets:0000000000000000 (System.UIntPtr)
    _safeSerializationManager:00000003800702c0 (System.Runtime.Serialization.SafeSerializationManager)

lanfeust69:
Just to add a bit : in Parser.cs, the beginning of the error message is skipped in

```
                    case OP_MINUS_ERR_SPC:
                        switch (b)
                        {
                            case ' ':
                            case '\t':
                                state = OP_MINUS_ERR_SPC;
                                break;
                            default:
                                state = MINUS_ERR_ARG;
                                break;
                        }
                        break;
```

where we transition to `MINUS_ERR_ARG` without using current `b` (which happens to be the `'`).

Looking at corresponding go code, it seems there is the same kind of problem (but I'm not familiar with go) :

```
        case OP_MINUS_ERR_SPC:
            switch b {
            case ' ', '\t':
                continue
            default:
                nc.ps.state = MINUS_ERR_ARG
                nc.ps.as = i
            }
        case MINUS_ERR_ARG:
            switch b {
            case '\r':
                nc.ps.drop = 1
            case '\n':
                var arg []byte
                if nc.ps.argBuf != nil {
                    arg = nc.ps.argBuf
                    nc.ps.argBuf = nil
                } else {
                    arg = buf[nc.ps.as : i-nc.ps.drop]
                }
                nc.processErr(string(arg))
                nc.ps.drop, nc.ps.as, nc.ps.state = 0, i+1, OP_START
            default:
                if nc.ps.argBuf != nil {
                    nc.ps.argBuf = append(nc.ps.argBuf, b)
                }
            }
```

where `arg` will include quotes, but `nc.ps.argBuf` (if present) will be missing the first one. Of course it is very likely that it never causes any actual problem...

As for the C#, I suppose it would be more consistent to keep the quotes, and either have the constant strings include them, or go through an equivalent of the go `normalizeErr()` function.

Also note that in all likelihood, the `STALE_CONNECTION` case isn't actually handled either.

ColinSullivan1:
@stephenjannin,  I think a few bugs were discovered and resolved from discussion around this PR.  Your input, along with @lanfeust69, has been incredibly valuable in shoring up some of the corner cases, and is very much appreciated!  After merging #74, do you feel this can be closed?

stephenjannin:
perfect, #74 will really fix the issue

PR #74, opened 2016-05-26
Fix connection closed on authorization timeout

lanfeust69:
This is an alternative to #68.

Also note that it would also make sense for the other nats client to have the same behavior (disconnect on authorization timeout, not close).

### DIFF
--- a/NATS/Conn.cs
+++ b/NATS/Conn.cs
@@ -1117,11 +1117,11 @@ namespace NATS.Client
                 else if (result.StartsWith(IC._ERR_OP_))
                 {
                     throw new NATSConnectionException(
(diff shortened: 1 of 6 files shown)


stephenjannin:
missing one unit test modification : UnitTestTLS.cs

ColinSullivan1:
@lanfeust69, good catch, thanks!  We're currently looking into this, and I'm leaning toward that this is contained to the .NET client.  Ideally, we should always handle an authorization error in the connection handshake since it is synchronous - perhaps the read loop is not shutting down immediately during reconnect.  That being said, this cleanly addresses this problem today.  Thanks for the contribution! LGTM.

PR #84, opened 2016-06-20
Removed unnecessary lock/capture of local variable

sixlettervariables:
Removes capture of local Parser variable in Connection's member field `ps`. The field `ps` was only ever assigned and never used.


ColinSullivan1:
@sixlettervariables This looks good...  I'm going to run a few stress tests against this before merging.

ColinSullivan1:
Apologies for the delay, the stress tests succeeded, LGTM.

PR #85, opened 2016-06-20
Saves exception context on rethrow in processExpectedInfo

sixlettervariables:
`Conn.processExpectedInfo` loses the stack trace via `throw e`, switched to a bare `throw` to rethrow with the correct context.


ColinSullivan1:

PR #86, opened 2016-07-16
Switched gitinore to official one from gitignore.io

cstlaurent:
Hi,

I just cloned the repo and noticed that files from Resharper are getting in the commits. I saw that the gitignore file was pretty slim so I changed it to the gitignore.io visual studio edition so that no unwanted files gets into the codebase.


ColinSullivan1:

Issue #62, opened 2016-05-02
Travis CI for the .NET client


ColinSullivan1:
It would be nice to have continuous integration for the .NET client.  Currently, Travis CI supports mono and a few test frameworks as beta.  The .NATS C# test suite uses the Visual Studio Test Suite to avoid external dependencies.

Investigate:
- Level of support (now and future) in Travis CI
- Other user experiences of .NET in Travis CI
- NUnit vs xunit, including level of effort to convert

cstlaurent:
Hi,

I don't know what is the progress on this one, but I think I could help with some pull requests if you are ok with it.

Here are my recommendations:
- **Test Framework**: I may be biased here since it's my framework of choice, but I think considering the fact that the CoreFx and ASP.NET teams chose XUnit as their official test framework it may provide more support in the future, switching from VS Tests to Xunit would be a good idea.
- **CI server**: I think Travis CI is not enough for a C# project since it only offers Mono on Linux support for now, and I'm pretty sure this is not the main deployment configuration for users. I think using AppVeyor would be a better idea for now since it offers CI builds on Windows machine with full .NET framework. Travis CI or maybe Circle CI could be used to test on Mono and .NET Core on Linux eventually.

Let me know what you think!

ColinSullivan1:
@cstlaurent Thank you for the recommendations!  I was planning on moving to xunit for exactly the reasons you mentioned, and in fact the new NATS streaming C# library (under development) will use xunit.  Any PRs you'd like to submit to move over to xunit would be greatly appreciated!

I agree with your note on the CI server, and will certainly look into AppVeyor.  However, since everything we use is already on Travis CI, I'd like to see how well it works first.

cstlaurent:
I think this issue could be closed after PR #87 and #89

PR #87, opened 2016-07-21
Convert tests to XUnit

cstlaurent:
## Goal of the pull request

Convert tests currently running on MSTests to XUnit to remove dependency on Visual Studio. As discussed in issue #62 
## What is needed
- [x] Convert MSTest project to regular class library since the .csproj itself fiddle with MSTest
- [x] Remove MSTest and add XUnit
- [x] Convert all Assert nomenclature to the XUnit one
- [x] Replace all  `if` checks with associated `Assert.Fail` in code since it does not exist in XUnit
- [x] Remove all `TestContext` in code since it does not exist in XUnit
- [x] Validate that all tests are still passing

### DIFF
--- a/NATS.sln
+++ b/NATS.sln
@@ -1,11 +1,9 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Express 2013 for Windows Desktop
-VisualStudioVersion = 12.0.31101.0
+# Visual Studio 14
+VisualStudioVersion = 14.0.25420.1
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NATS", "NATS\NATS.csproj", "{68EE71D4-9532-470E-B5CA-ECAA79936B1F}"
-EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NATSUnitTests", "NATSUnitTests\NATSUnitTests.csproj", "{00DBFD4D-72F4-4250-884C-C1527C66A0C2}"
 EndProject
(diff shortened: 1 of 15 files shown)


cstlaurent:
Hi, I found out a bug in the tests while converting it to XUnit. Actually multiple instance of it...

The bug is in `TestAsyncSubscriptionPending()` inside the `TestSubscriptions` class, line 561 (in the MSTest codebase)

The bug is a testing technique that was used throughout the test code. It's actually a combination of `Assert.Fail()` inside `try-catch` blocks where the Assert.Fail is placed inside the try block after a line of code that should fail. This was to check that the code under test actually throws an exception, as says the Assert message: `"Should have received an exception."`.

The problem is represented in the following screenshot:
image

Here `s.ClearMaxPending();` should throw an exception as intended the test, but it does not, so the `Assert.Fail();` execute, which throws an exception that is swallowed by the empty catch block.

So basically the test was passing not because `s.ClearMaxPending();` was throwing an exception but because the Assert exception was swallowed by a catch block.

And this technique was used all over the place. It's a chance that this is the only failing test after the conversion to XUnit.

cstlaurent:
If I comment out the 3 last Exception checks, the test pass:
- `s.ClearMaxPending()`
- `s.Delivered`
- `s.Dropped`

This is the last thing remaining to finalize this pull-request. Should I leave it this way and have the pull-request accepted with 1 failing test? Or should I comment the 3 failing checks to make the test pass?

ColinSullivan1:
@cstlaurent, I really appreciate the work you've put into this contribution!  Good catch on the tests there - to finalize this you should be able to actually remove the three failing exception tests (those APIs do not throw exceptions - at one point in early development they did).  I'll do some compatibility testing with the project file in the meantime.  Great work!

cstlaurent:
@ColinSullivan1, I removed the invalid checks. I think the pull request is pretty much done on my side at this point.

Thanks!

PR #88, opened 2016-07-22
Remove interpolated strings

ColinSullivan1:
No functional changes; just not quite ready to require C# 6 features in the source code to support legacy users.

### DIFF
--- a/NATSUnitTests/UnitTestBasic.cs
+++ b/NATSUnitTests/UnitTestBasic.cs
@@ -36,7 +36,7 @@ namespace NATSUnitTests
 
             string u = c.ConnectedUrl;
 
-            Assert.False(string.IsNullOrWhiteSpace(u), $"Invalid connected url {u}.");
+            Assert.False(string.IsNullOrWhiteSpace(u), string.Format("Invalid connected url {0}.", u));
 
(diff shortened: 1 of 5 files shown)


cstlaurent:
I don't want to argue but I'm not sure I understand your point 100%, since currently NATS is compiled to .NET 4.5 which has full C# 6 support. Also like I said, C# 6 is a compiler only "thing" and most of it can be used and run in older .NET code base. String interpolation, for example, is compatible up to .NET 3.5 like most of the C# 6 features. You get compiler errors if this feature is not supported on the targetted runtime framework.

Also, all Microsoft supported version of Windows should have 4.5.2 framework at least by now since it was pushed through Windows updates a long time ago.

ColinSullivan1:
I do see your point.  We do have at least one user who's deployed NATS into production who at the time required a 4.0 compatibly source code base due to their restrictions - they may wish to remain anonymous, but it's related to https://github.com/nats-io/csnats/issues/49.  I think it'd be sufficient for most users (or the NATS team) to compile down to the compatibility level needed, and that's something to look at in the future.  I'll poll the user community to see if we can remove this restriction in the next release.

PR #89, opened 2016-07-23
CI support

cstlaurent:
Hi @derekcollison 

I added basic support for CI builds on AppVeyor and Travis-CI. 

For AppVeyor, I created the configuration files `appveyor.yml` with everything needed to build the solution and run all unit tests. The scripts download the latest version of gnatsd.exe from the GitHub repo so that the tests can run. The solution builds correctly and tests results are available online, you can see it here: [build](https://ci.appveyor.com/project/cstlaurent/csnats) and here [tests](https://ci.appveyor.com/project/cstlaurent/csnats/build/tests). There seem to be some flaky tests that sometimes fail, depending on the order they runs I imagine...

On the other hand, for Travis-CI, I added a basic config file but it looks like the solution does not build on a mono target with some errors related to encryption, so I left it at that: 
image

You can see my Travis Build [here](https://travis-ci.org/cstlaurent/csnats)

I think the build script is a bit hacky right now and a lot more could be added but I think it's a pretty good start!

All you would have to do now is to create accounts on Travis and AppVeyor (if you don't have one already), enable builds on the csnats repo and everything should work properly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,2 @@
+language: csharp
+solution: NATS.sln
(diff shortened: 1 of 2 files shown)


ColinSullivan1:
@cstlaurent, Thank you for doing this.  I did look into AppVeyor, and it does look to be a good way to go.  I was heading in that direction for the upcoming C# Streaming client.  @derekcollison, I'll look at getting accounts setup if you are OK with using AppVeyor.

ColinSullivan1:
Just adding a note; this issue is related to #62.

Issue #91, opened 2016-07-26
User references should be released after a NATS object has been disposed


ColinSullivan1:
Some user references remain in NATS objects after they are disposed, namely event handlers.  Look at all NATS objects for these occurrences, and release any references to user objects during Dispose.

ColinSullivan1:
#94 partially addresses this issue.  Further tests need to be done with respect to subscriptions.

PR #143, opened 2016-12-28
Memory leak fix

ogoun:
Remove CallbackScheduler, replaced on Task.Factory seeing memory leak.
Memory leak was observed when the server is turned off and attempt to reconnect to it.

### DIFF
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -32,7 +32,7 @@ namespace NATS.Client
     internal enum ClientProtcolVersion
     {
         // clientProtoZero is the original client protocol from 2009.
- 	    // http://nats.io/documentation/internals/nats-protocol/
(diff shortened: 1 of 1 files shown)


ColinSullivan1:
@ogoun , Thank you for the contribution, much appreciated!  Did you identify which objects were causing the memory leak?

The home grown task scheduler was put in place to ensure the ordering of tasks put into it (e.g. it wouldn't make sense to get a reconnected event handler occur before a disconnected event handler). Unfortunately, creating tasks with the default task scheduler can't guarantee execution order.

ColinSullivan1:
We really appreciate you contributing!  Just figured I'd check again - do you know which objects were leaking, or would you have a test to reproduce this?  Thanks!

ColinSullivan1:
Closing due to inactivity, but keep this in mind when addressing #91.

Issue #159, opened 2017-05-30
Add handler invoked when a new server joins the NATS Cluster 


ColinSullivan1:
Add functionality to mirror the go-nats feature,https://github.com/nats-io/go-nats/pull/282.

PR #160, opened 2017-06-05
Implements ServerDiscoveredEventHandler from #159

sixlettervariables:
ServerPool.add matches its description and returns a boolean indicating if the server url
is new for the pool. Connection.processInfoArgs uses this to signal on the new event if and
only if it has been asked to notify on server discovery. The initial connect path does not enable
that behavior, but the processAsyncInfo path does enable it, matching the behavior of the
golang client https://github.com/nats-io/go-nats/pull/282

### DIFF
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -975,6 +975,7 @@ namespace NATS.Client
 
         // processExpectedInfo will look for the expected first INFO message
         // sent when a connection is established. The lock should be held entering.
+        // Caller must lock.
         private void processExpectedInfo()
         {
             Control c;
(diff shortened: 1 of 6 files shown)


ColinSullivan1:
Thank you for the PR!  Much appreciated!

It looks like the the test is failing in AppVeyor though...  https://ci.appveyor.com/project/NATS-CI47222/csnats/build/1.0.296/job/4568i53hvd715jpr/tests

sixlettervariables:
Interesting, passes locally. I wonder if the 5 seconds I wait for the cluster to converge is not long enough on AppVeyor.

ColinSullivan1:
@sixlettervariables , Most likely not, AppVeyor can be very resource constrained.  The silver lining is that it is good for finding race conditions.  Can you set it to 30 seconds and give it a shot?

ColinSullivan1:
...Perfect.  Let's see if it works on AppVeyor.

ColinSullivan1:
I think I see the problem. The server version should be set 0.9.6.  I'll create a separate PR.

ColinSullivan1:
LGTM!  There's a flapper in CI unrelated to this change.  Thanks for the contribution!

Issue #158, opened 2017-05-30
More protocol friendly request pattern. 


ColinSullivan1:
Add functionality to mirror the go-nats feature, https://github.com/nats-io/go-nats/pull/295.

sixlettervariables:
If I'm reading this correctly, does the following describe that behavior?

- Each client (when initiating the first request) would create a connection local inbox
- The client (read: per connection) would subscribe to that inbox as a wildcard
- Each request from the client would have a reply to referencing that inbox plus a unique ID
- The reply would be dispatched locally based on that unique ID

ColinSullivan1:
@sixlettervariables , were you thinking of contributing here?

sixlettervariables:
As I understand it we, basically, do exactly this in our usage of the client (but for a different purpose than performance). Trying to understand the feature and how it may/may not affect us.

ColinSullivan1:
It really shouldn't affect you (beyond better performance), unless you have some subject based authorization enabled that would interfere with the internal subscription.

ColinSullivan1:
Resolved by #167.

PR #161, opened 2017-06-08
Fix weak reply name for Request() call.

alexanderinochkin:
Current implementation of NewInbox() could generate identical names when several Request() calls could be done from different threads at the close time.
The problem is that Random() uses Environment.TickCount as a seed. When program makes several Request() calls in the same system tick (500 ms), the several Random().NextBytes() calls will generate identical inbox names for reply for different requests. In that case both requests will get the first server response.
To solve it I suggest to use more strongly generator like Guid.NextGuid() for inbox name.

### DIFF
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -53,8 +53,6 @@ namespace NATS.Client
         // .NET 4.0.
         readonly internal object mu = new Object(); 
 
-        private Random r = null;
-
         Options opts = new Options();
 
         // returns the options used to create this connection.
@@ -2099,14 +2097,7 @@ namespace NATS.Client
 
         public string NewInbox()
         {
-            if (r == null)
-                r = new Random();
-
-            byte[] buf = new byte[13];
-
-            r.NextBytes(buf);
-
-            return IC.inboxPrefix + BitConverter.ToString(buf).Replace("-","");
+            return IC.inboxPrefix + Guid.NewGuid().ToString("N").ToUpperInvariant();
         }
 
         internal void sendSubscriptionMessage(AsyncSubscription s)
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -51,7 +51,9 @@ namespace NATS.Client
 
         // NOTE: We aren't using Mutex here to support enterprises using
         // .NET 4.0.
-        readonly internal object mu = new Object(); 
+        readonly internal object mu = new Object();
+
+		private Random r = null;
 
         Options opts = new Options();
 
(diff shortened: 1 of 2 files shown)


sixlettervariables:
Perhaps a better approach would be to use a hash value of a new GUID as the seed? This would enable using the higher performance of the `Random` without the expense of a new GUID creation on every request/reply.

Conceptually, if #158 gets implemented using a GUID as the inbox (or something with similar uniqueness) then that would solve this problem.

alexanderinochkin:
Let me compare current implementation and the new one by performance:
```
var tr = Stopwatch.StartNew();
var r = new Random();
for (int i = 0; i < 1000000; i++)
{
	byte[] buf = new byte[13];
	r.NextBytes(buf);
	var bits = BitConverter.ToString(buf).Replace("-", "");
}
Console.WriteLine($"Elapsed Random(): {tr.Elapsed}");

var tg = Stopwatch.StartNew();
for (int i = 0; i < 1000000; i++)
{
	var guid = Guid.NewGuid().ToString("N").ToUpperInvariant();
}
Console.WriteLine($"Elapsed GUID:     {tg.Elapsed}");
```
The result is: 
```
Elapsed Random(): 00:00:00.7047381
Elapsed GUID:     00:00:00.6231147
```
If uppercased name is not required, then it is possible to omit ToUpperInvariant(). Finally the "guid used" implementation will be just about twice faster then the current one:
```
Elapsed Random(): 00:00:00.7164708
Elapsed GUID:     00:00:00.4075821
```

sixlettervariables:
Good timing information to have, but I saw the opposite on my laptop (interestingly enough). Of note, I made the following change to the loop to avoid throwing away the result of the operation:
```csharp
string[] keepThisValue = new string[2];
// ...
    keepThisValue[0] = BitConverter.ToString(buf).Replace("-", "");
// ...
    keepThisValue[1] = Guid.NewGuid().ToString("N").ToUpperInvariant();
```

Giving me:
```
Elapsed Random(): 00:00:01.0181849
Elapsed GUID:     00:00:02.9218345
```

alexanderinochkin:
Your test give me the next results (i7-6700 CPU 3.4GHz Win10):
```
Elapsed Random(): 00:00:00.7384288
Elapsed GUID:     00:00:00.5470446
```

sixlettervariables:
It appears that my old MacBook Pro (Core 2 Duo) is showing its age!

alexanderinochkin:
Christopher , I've changed the code as you suggested.
I hope this will ensure the uniqueness of names at high performance on every platform.

ColinSullivan1:
@alexanderinochkin , the code looks good to me; I'm glad you've caught this.

Just a few things...   Can you add a test for this that fails with the current code to ensure it is fixed?  Also, it looks like tabs/alignment have been changed.  If you could remove the whitespace changes that'd be great.

Your contribution is much appreciated!

alexanderinochkin:
@ColinSullivan1 , the test is added to the pull request.

ColinSullivan1 (approved):

PR #162, opened 2017-06-08
Fix problem with parsing server url.

alexanderinochkin:
When passed urlString starts with ‘nats’ (like ‘nats.domain.com’ or 'natsserver.company.com') the protocol supplement will be skipped and connection will fail.


ColinSullivan1:

PR #167, opened 2017-06-21
Add Single Request subscription mode

sixlettervariables:
This implements #158 by default, and provides `Options.UseOldRequestStyle` to change the behavior. The implementation mirrors go-nats, with differences expected for the differences in framework/language.

- Adds an InFlightRequest object to track each request, as well as provide a common cancellation token source.
- Uses a GUID for the inbox (per #161)
- Does not change the public API for requests, instead choosing to branch based on the chosen request model
- Unit tests expanded to test both old and new strategies
- Resolves intermittent failure behavior in Task based Request-Reply unit tests
- Adds Async request-reply to the benchmark suite

### DIFF
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -2252,8 +2252,15 @@ namespace NATS.Client
             {
                 var request = requestSync(subject, data, timeout, CancellationToken.None);
 
-                Flush(timeout > 0 ? timeout : DEFAULT_FLUSH_TIMEOUT);
-                request.Waiter.Task.Wait(timeout);
+                try
+                {
+                    Flush(timeout > 0 ? timeout : DEFAULT_FLUSH_TIMEOUT);
+                    request.Waiter.Task.Wait(timeout);
+                }
+                catch (NATSTimeoutException)
+                {
+                    removeTimedoutRequest(request.Id);
+                }
 
                 return request.Waiter.Task.Result;
             }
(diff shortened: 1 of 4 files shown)


NATS.Client/Conn.cs
```diff
@@ -100,6 +101,62 @@ public Options Opts
...
+
+                        timeoutToken.Token.Register(
+                            () => this.Waiter.TrySetException(new NATSTimeoutException()));
+                        timeoutToken.CancelAfter(timeout);
+                    }
+                }
+            }
+
+            public string Id { get; set; }
+            public CancellationToken Token { get; }
```
ColinSullivan1: Could you add `set;` to `CancellationToken` and  `TaskCompletionSource`? To date, we've been maintaining C# 5 compatibility (will move to C# 6 fairly soon, though).

ColinSullivan1 (changes requested):
@sixlettervariables , Overall, it looks great, thanks!  Just one change requested.

NATS.Client/Conn.cs
```diff
@@ -2191,6 +2248,165 @@ public void Publish(string subject, string reply, byte[] data)
...
+                var request = requestSync(subject, data, timeout, CancellationToken.None);
+
+                try
+                {
+                    Flush(timeout > 0 ? timeout : DEFAULT_FLUSH_TIMEOUT);
+                    request.Waiter.Task.Wait(timeout);
+                }
+                catch (NATSTimeoutException)
+                {
+                    removeTimedoutRequest(request.Id);
```
sixlettervariables, line 2262: Missed a throw.

ColinSullivan1:
LGTM!

PR #169, opened 2017-06-29
Implement negative values for unlimited pending limits

### LINKED ISSUE #168: Negative values for pending limits are not treated the same as the Go client
The documentation for the Go client suggests that the pending message and byte limits treat negative values as "no limits", however the C# client does not contain this logic.

sixlettervariables:
This fixes #168 and implements the logic for unlimited pending bytes and messages. Adds documentation for the same as well (to match the Go client).


ColinSullivan1 (approved):
LGTM.  Much Appreciated!

Issue #171, opened 2017-07-11
I need a publish method where I can pass in the byte count


BrannonKing:
I would like to be able to pass in the byte count that goes with my byte[] passed to the Publish method. In other words, I want to set the "num" value in Connection.publish with a parameter so that I don't have to make an additional array copy of my data after serialization.

sixlettervariables:
Would you also want an offset parameter?

Something like:
```
void Publish(string subject, byte[] data, int count);
void Publish(string subject, byte[] data, int offset, int count);
void Publish(string subject, string reply, byte[] data, int count);
void Publish(string subject, string reply, byte[] data, int offset, int count);
```

BrannonKing:
I don't need "offset" in my current application, but there have been some in times past where it would be handy. Those overloads you listed look fine.

ColinSullivan1:
I think `Publish` APIs with offset and count would keep the API minimized while also handling a wider use case.  For congruence, the `Request*` APIs could provide offsets and count as well.

aricart:
@BrannonKing this is not needed if your 'array' is actually a slice.
If you have a 'slice', you can subrange the slice, the backing of the slice remains to be the original slice. 

So all you need to do is subrange your source slice, and the api will send the subset of bytes, without any data copying. Have a look at this sample:
https://play.golang.org/p/XHs7KlPm78

BrannonKing:
C# doesn't support native array subranges. In order to sub-range a collection in C#, the API would still have to change. If it took a `IReadOnlyList<byte>`, it's easy to build a sub-range wrapper around around that interface. Another option is to take `ArraySegment<byte>`, which bundles the Count and Offset parameters with the array.

aricart:
bah! sorry thought it was go-nats!

sixlettervariables:
vNext of the CLR will have [`Span<T>`](https://github.com/dotnet/coreclr/issues/5851) (rough analog of go slices).

BrannonKing:
@sixlettervariables , I see that you prototyped this. What's the status on a merge?

sixlettervariables:
@BrannonKing I temporarily stepped onto another project at work and just never pushed my final changes. I've opened a PR.

ColinSullivan1:
Resolved by #182.

PR #182, opened 2017-09-22
Adds offset+count overloads to Publish and Request

sixlettervariables:
Implements #171.

### DIFF
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -2245,19 +2245,29 @@ namespace NATS.Client
         // publish is the internal function to publish messages to a nats-server.
         // Sends a protocol data message by queueing into the bufio writer
         // and kicking the flush go routine. These writes should be protected.
-        internal void publish(string subject, string reply, byte[] data)
+        internal void publish(string subject, string reply, byte[] data, int offset, int count)
         {
             if (string.IsNullOrWhiteSpace(subject))
             {
                 throw new NATSBadSubscriptionException();
             }
-
-            int msgSize = data != null ? data.Length : 0;
+            else if (offset < 0)
+            {
+                throw new ArgumentOutOfRangeException("offset");
+            }
+            else if (count < 0)
+            {
+                throw new ArgumentOutOfRangeException("count");
+            }
+            else if (data != null && data.Length - offset < count)
+            {
+                throw new ArgumentException("Invalid offset and count for supplied data");
+            }
 
             lock (mu)
             {
                 // Proactively reject payloads over the threshold set by server.
-                if (msgSize > info.maxPayload)
+                if (count > info.maxPayload)
                     throw new NATSMaxPayloadException();
 
                 if (isClosed())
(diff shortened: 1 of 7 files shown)


ColinSullivan1:
@sixlettervariables Thanks! (as always).  I'll starting reviewing this today.

NATS.Client/Conn.cs
```diff
@@ -2330,7 +2357,8 @@ public void Publish(Msg msg)
                 throw new ArgumentNullException("msg");
             }
 
-            publish(msg.Subject, msg.Reply, msg.Data);
+            int count = msg.Data != null ? msg.Data.Length : 0;
+            publish(msg.Subject, msg.Reply, msg.Data, 0, count);
```
ColinSullivan1: Could use the expression above and avoid declaring count.  This is where I wish C# had macros.

NATS.Client/Conn.cs
```diff
@@ -2351,14 +2379,54 @@ public void Publish(Msg msg)
...
+            publish(subject, reply, data, offset, count);
+        }
+
+        internal virtual Msg request(string subject, byte[] data, int offset, int count, int timeout)
+        {
+            if (string.IsNullOrWhiteSpace(subject))
+            {
+                throw new NATSBadSubscriptionException();
+            }
+            else if (offset < 0)
```
ColinSullivan1: Maybe a comment instead of these arg checks?  `offset`, `count`, and the overrun will be checked by the `publish` method later on.   It'll also keep the arg checking logic in one place.

NATS.Client/Conn.cs
```diff
@@ -2440,34 +2508,63 @@ private InFlightRequest requestSync(string subject, byte[] data, int timeout, Ca
...
-                async () =>
-                {
-                    InFlightRequest request = setupRequest(timeout, token);
-
-                    request.Token.ThrowIfCancellationRequested();
+            if (string.IsNullOrWhiteSpace(subject))
+            {
+                throw new NATSBadSubscriptionException();
+            }
+            else if (offset < 0)
```
ColinSullivan1: Same as above regarding argument checking.

NATS.Client/Conn.cs
```diff
@@ -2560,15 +2657,34 @@ private Msg oldRequest(string subject, byte[] data, int timeout)
...
+        /// to the connected NATS server.</param>
+        /// <param name="offset">The zero-based byte offset in <paramref name="data"/> at which to begin publishing
+        /// bytes to the subject.</param>
+        /// <param name="count">The number of bytes to be published to the subject.</param>
+        /// <param name="timeout">The number of milliseconds to wait.</param>
+        /// <returns>A <see cref="Msg"/> with the response from the NATS server.</returns>
+        /// <seealso cref="IConnection.Request(string, byte[])"/>
+        public Msg Request(string subject, byte[] data, int offset, int count, int timeout)
+        {
+
```
ColinSullivan1: Total nitpick, but there's an extra line here.

NATS.Client/SyncSub.cs
```diff
@@ -104,7 +104,7 @@ public Msg NextMessage(int timeout)
                 {
                     d = tallyDeliveredMessage(msg);
                 }
-                if (d == max)
+                if (d == localMax)
```
ColinSullivan1, line 107: Thanks, good catch!

NATS.Client/Conn.cs
```diff
@@ -2514,15 +2611,15 @@ private void removeTimedoutRequest(string requestId)
             }
         }
 
-        private Msg oldRequest(string subject, byte[] data, int timeout)
+        private Msg oldRequest(string subject, byte[] data, int offset, int count, int timeout)
```
ColinSullivan1, line 2591: This will eventually be deprecated, but thanks for adding the offset/count to these APIs.

ColinSullivan1 (changes requested):
Thanks for the contribution!  Overall LGTM.  Just one concern about duplicate arg checking.

NATS.Client/Conn.cs
sixlettervariables: Done.

NATS.Client/Conn.cs
sixlettervariables: Done.

NATS.Client/Conn.cs
sixlettervariables: Honestly don't know how I missed that. I nitpick the mess out of whitespacing!

ColinSullivan1 (approved):
LGTM - thanks!

PR #185, opened 2017-10-10
Apply timeout option when connect to NATS server

alexanderinochkin:
This change is to be able to use timeout option when connect to NATS server. Timeout option is not applied in NET 4.5 version of NATS client  and default timeout (about 80 sec) is used.

### DIFF
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -388,15 +388,12 @@ namespace NATS.Client
                         sslStream = null;
                     }
 
-#if NET45
-                    client = new TcpClient(s.url.Host, s.url.Port);
-#else
                     client = new TcpClient();
                     if (!client.ConnectAsync(s.url.Host, s.url.Port).Wait(TimeSpan.FromMilliseconds(timeoutMillis)))
                     {
+                        client = null;
                         throw new NATSConnectionException("timeout");
                     }
-#endif
 
                     client.NoDelay = false;
 


ColinSullivan1:
@alexanderinochkin , thank you for the PR, I'll take a look.

ColinSullivan1:

sixlettervariables:
For folks in the future who may be searching for why they are now receiving NATS Timeouts, this fix led to a regression in part of our application due to an invalid assumption on our part. We discovered this assumption while rolling back to 0.8.0 during debugging.

Our web services layer uses IIS, which hosts some Web API components that fire messages over NATS. In the latest release of our application we began seeing failures during dependency injection of NATS, stating that the connection had timed out. We initially struggled to locate the issue, as other parts of the application using NATS did not have any issues or timeouts. Further debugging located IIS App Pool recycles as the trigger to cause the problem.

We believe during a cold start the large number of threads spinning up as our clients make requests causes an initial connection to NATS to be delayed. This delay under 0.8.0 would fall within the default Windows timeout of ~80-90 seconds. Under 0.8.1 this would now fall outside the default NATS C# Client timeout of 2 seconds. We had not configured this timeout previously, as we erroneously believed the default was sufficient, not realizing the default was not actually being applied.

ColinSullivan1:
@sixlettervariables , do you need us to revert back?

sixlettervariables:
@ColinSullivan1 no, we're good (thank you though). I missed the details of this bug fix and never tested to see if the "default" timeout actually worked for us. The rollback was painless and the fix (setting the timeout) is already in our next version.

PR #186, opened 2017-10-11
prevent pingTimer from exhausting threads

stephenjannin:
Hello,

It's a long time without issues for us !

if doReconnect thread is slow or blocked (owning the mu lock), then the ping timer keeps creating threads waiting for the mu lock, until resources are exhausted.

We have a memory dump with > 900 threads for this timer. I'm still trying to find why the doReconnect thread was blocked while reading from the underlying socket

### DIFF
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -877,6 +877,11 @@ namespace NATS.Client
                     sendPing(null);
                 }
                 catch (Exception) { }
+                //trigger timer again
+                if (ptmr != null)
+                {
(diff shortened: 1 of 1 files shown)


sixlettervariables:
This certainly seems like the desired behavior: only one in-flight PING. Subsequent PINGs would not even be scheduled until after the current one succeeds.

ColinSullivan1:
@stephenjannin , Thanks - much appreciated!  LGTM.  Once I test this manually (AppVeyor seems to be having issues), I'll merge it.

ColinSullivan1:
Tests pass.  Thanks again!

Issue #157, opened 2017-05-05
Add ManyRequests feature


ogoun:
From official documentation:

> NATS supports two flavors of request reply messaging: point-to-point or **one-to-many**. Point-to-point involves the fastest or first to respond. In a one-to-many exchange, you set a limit on the number of responses the requestor may receive.

Your implementation does not contain a method for one-to-many scenario.

**Implementation example**

```
public IEnumerable<Msg> RequestMany(string subject, byte[] data, int timeout)
{
    if (timeout <= 0)
    {
        throw new ArgumentException("Timeout must be greater that 0.", "timeout");
    }
    string inbox = InboxNameGen.NewInbox();
    SyncSubscription s = subscribeSync(inbox, null);
    publish(subject, inbox, data);
    Flush();
    var set = new List<Msg>();
    Msg m = null;
    do
    {
        try
        {
            m = s.NextMessage(timeout);
            set.Add(m);
        }
        catch(Exception ex)
        {
            break;
        }
    } while (m != null);
    s.unsubscribe(false);
    return set;
}
```

ColinSullivan1:
@ogoun , Thank you for the suggestion, it is very much appreciated.  This topic has come up in the past; we'll consider this.  Note, if something like this is implemented, we will likely start with the [golang](https://github.com/nats-io/go-nats) client, then we'll add to other clients (including C# here).

PR #187, opened 2017-10-12
Always receive server connection notices

madgrenadier:
- [x] You have certified that the contribution is your original work and that you license the work to the project under the [MIT license](https://github.com/nats-io/gnatsd/blob/master/LICENSE)

### Changes proposed in this pull request: 

In conjunction with PR #597, here is the csharp client code changes that enable the INFO message to always pass through to the event handlers.  In this case, the event handler named 'ServerDiscoveredEventHandler' is allowed to trigger regardless if the client server list already has that server node in its list.  PR #597 allows the server to always send notification, and this change allows the client to always receive the change event notice.

Consequently, when combined with the server changes in PR#597, this allows user application code to be triggered every time there is a topology change to the clusters.  It will be up to the user application code to handle the event and maintain a list of server URLs to perform the desired (if any) reconnect.

### Who Benefits From The Change(s)?
Any configuration where the user application might want to be notified of a topology change regardless of cause (a reconnecting server, in this case) would benefit from this change.

### DIFF
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -1057,6 +1057,8 @@ namespace NATS.Client
                     {
                         processConnectInit();
                         exToThrow = null;
+
+                        Console.WriteLine("Connected to: {0}", s.url);
                         return true;
                     }
                 }
@@ -1147,7 +1149,7 @@ namespace NATS.Client
             }
 
             // do not notify listeners of server changes when we process the first INFO message
-            processInfo(c.args, false);
+            processInfo(c.args, false, false);
             checkForSecure();
         }
 
@@ -1512,6 +1514,8 @@ namespace NATS.Client
                 }
                 catch (Exception) { }
 
+                Console.WriteLine("Successfully Reconnected to: {0}", cur.url);
+
                 return;
             }
 
@@ -1569,6 +1573,17 @@ namespace NATS.Client
                 try
                 {
                     len = br.Read(buffer, 0, Defaults.defaultReadLength);
+
+                    if ( len == 0 )
+                    {
+                        Console.WriteLine("Zero length message encountered");
+                        throw new NATSException("Invalid Message length");
+                        // continue;
+                    }
+
+                    Console.WriteLine("=======================================");
+                    Console.WriteLine("IN ({0}): {1}", len, Encoding.Default.GetString(buffer, 0, len));
+
                     parser.parse(buffer, len);
                 }
                 catch (Exception e)
@@ -2112,7 +2127,7 @@ namespace NATS.Client
         // processInfo is used to parse the info messages sent
         // from the server.
         // Caller must lock.
-        internal void processInfo(string json, bool notifyOnServerAddition)
+        internal void processInfo(string json, bool notifyOnServerAddition, bool forced)
         {
             if (json == null || IC._EMPTY_.Equals(json))
             {
@@ -2134,7 +2149,7 @@ namespace NATS.Client
                 }
 
                 var serverAdded = srvPool.Add(servers, true);
-                if (notifyOnServerAddition && serverAdded)
+                if (notifyOnServerAddition && (serverAdded || forced))
                 {
                     scheduleConnEvent(opts.ServerDiscoveredEventHandler);
                 }
@@ -2145,7 +2160,7 @@ namespace NATS.Client
         {
             lock (mu)
             {
-                processInfo(Encoding.UTF8.GetString(jsonBytes, 0, length), true);
+                processInfo(Encoding.UTF8.GetString(jsonBytes, 0, length), true, opts.AlwaysReconnect);
             }
         }
 
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -1058,7 +1058,6 @@ namespace NATS.Client
                         processConnectInit();
                         exToThrow = null;
 
-                        Console.WriteLine("Connected to: {0}", s.url);
                         return true;
                     }
                 }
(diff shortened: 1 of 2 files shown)


madgrenadier:
I have modified Publish and Subscribe examples which can utilize this.

TheVons:
What are the chances of this being implemented?

NATS.Client/Conn.cs
```diff
@@ -1569,6 +1573,17 @@ private void readLoop()
                 try
                 {
                     len = br.Read(buffer, 0, Defaults.defaultReadLength);
+
+                    if ( len == 0 )
```
ColinSullivan1, line 1574: Did you encounter an error here?

madgrenadier: Yes, I was seeing this zero length reply happen when I was testing repeated disconnects and reconnects using the csharp client code with the publish/subscribe example pair.

I do not have the details handy for a repro at the moment, but I'll see if I can set it up again sometime soon.

Now, this was against 1.0.2 (I think) server, so that may no longer be the case, as I have not tested the same scenario against the more recent 1.0.4 server, so it may no longer be an issue.

NATS.Client/Conn.cs
```diff
@@ -1057,6 +1057,8 @@ internal bool connect(Srv s, out Exception exToThrow)
                     {
                         processConnectInit();
                         exToThrow = null;
+
+                        Console.WriteLine("Connected to: {0}", s.url);
```
sixlettervariables: Remove.

NATS.Client/Conn.cs
```diff
@@ -1512,6 +1514,8 @@ private void doReconnect()
                 }
                 catch (Exception) { }
 
+                Console.WriteLine("Successfully Reconnected to: {0}", cur.url);
```
sixlettervariables: Remove.

NATS.Client/Conn.cs
```diff
@@ -1569,6 +1573,17 @@ private void readLoop()
...
                     len = br.Read(buffer, 0, Defaults.defaultReadLength);
+
+                    if ( len == 0 )
+                    {
+                        Console.WriteLine("Zero length message encountered");
+                        throw new NATSException("Invalid Message length");
+                        // continue;
+                    }
+
+                    Console.WriteLine("=======================================");
```
sixlettervariables: Remove.

NATS.Client/Options.cs
```diff
@@ -330,6 +331,16 @@ public string Token
         }
 
         /// <summary>
+        /// Gets or sets the flag to control whether we always reconnect
+        /// on a new server entering the mesh
+        /// </summary>
+        public bool AlwaysReconnect
```
sixlettervariables: This property does not actually control reconnects, but rather notifications. I suggest renaming this to better reflect its intended effect.

NATS.Client/Conn.cs
```diff
@@ -2112,7 +2127,7 @@ internal void processOK()
         // processInfo is used to parse the info messages sent
         // from the server.
         // Caller must lock.
-        internal void processInfo(string json, bool notifyOnServerAddition)
+        internal void processInfo(string json, bool notifyOnServerAddition, bool forced)
```
sixlettervariables: `forced` does not quite explain the effect of this argument. Perhaps: `notifyOnServerReconnects`? Not 100% certain on a good name.

ColinSullivan1:
@madgrenadier , thank you for the PR. 

I do agree with @sixlettervariables comments.

Because this is a change that affects the server and all Apcera supported clients, let's keep this open, but track discussion of the feature on the gnatsd repo:  https://github.com/nats-io/gnatsd/pull/597.

NATS.Client/Conn.cs
madgrenadier: Agreed

NATS.Client/Conn.cs
madgrenadier: Agreed

NATS.Client/Conn.cs
madgrenadier: Agreed

NATS.Client/Conn.cs
madgrenadier: I like the suggested name.  Will change.

NATS.Client/Options.cs
madgrenadier: Agreed.  Will change the property name to AlwaysNotifyOnReconnect

kozlovic: Again, this is not correct. Your changes are simply notifying the registered callback that the topology changes (or not really since it is an old server coming back on), there is no reconnect on the client side here. The comment is wrong, so is the name.

madgrenadier: Would the property name '*AlwaysAllowServerAdditionNotification*', with associated comment:

```
 /// Gets or sets the flag to control whether to always allow
 /// the ServerDiscoveredEventHandler to run if the notifyOnServerAddition
 /// flag is set.
```

be acceptable?

madgrenadier:
Okay, well here is the cleaned up change set and I'll keep an eye on this PR and the server PR for ya'll guys' comments.

Thanks.

ColinSullivan1:
Thanks @madgrenadier !

TheVons:
What is the status of this?

ColinSullivan1:
@Vonnguru , please refer to https://github.com/nats-io/gnatsd/pull/597.  We will get back to you on this.

ColinSullivan1:
Because PR #157 in the NATS server was closed, I'm closing this PR.  Comments are here: https://github.com/nats-io/nats-server/pull/597#issuecomment-344432132.  We do really appreciate the effort that went into this.

PR #188, opened 2017-11-06
remove unnecessary lock during processMsg

tamnm:
Lock during conn.processMsg is unnecessary. It's single thread processing, already thread safe.

### DIFF
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -1939,35 +1939,27 @@ namespace NATS.Client
             bool maxReached = false;
             Subscription s;
 
-            lock (mu)
+            stats.inMsgs++;
+            stats.inBytes += length;
+
+            // In regular message processing, the key should be present,
+            // so optimize by using an an exception to handle a missing key.
+            // (as opposed to checking with Contains or TryGetValue)
+            try
             {
-                stats.inMsgs++;
-                stats.inBytes += length;
+                s = subs[msgArgs.sid];
+            }
+            catch (Exception)
+            {
+                // this can happen when a subscriber is unsubscribing.
+                return;
+            }
 
-                // In regular message processing, the key should be present,
-                // so optimize by using an an exception to handle a missing key.
-                // (as opposed to checking with Contains or TryGetValue)
-                try
-                {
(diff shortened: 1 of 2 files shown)


NATS.Client/Conn.cs
```diff
@@ -1939,35 +1939,27 @@ internal void processMsg(byte[] msg, long length)
...
-                        s.addMessage(new Msg(msgArgs, s, msg, length), opts.subChanLen);
-                    } // maxreached == false
-
-                } // lock s.mu
+            // In regular message processing, the key should be present,
+            // so optimize by using an an exception to handle a missing key.
+            // (as opposed to checking with Contains or TryGetValue)
+            try
+            {
+                s = subs[msgArgs.sid];
```
ColinSullivan1: `subs` can be accessed through unsubscribe outside of the readLoop.  While subs itself is in a thread-safe collection, we shouldn't allow processing on a subscription while it is being unsubscribed.

NATS.Client/Subscription.cs
```diff
@@ -109,15 +109,11 @@ public Connection Connection
  
         internal bool tallyMessage(long bytes)
```
ColinSullivan1: Good catch - this looks to be an unnecessary lock.  Would you mind adding a comment above the method (line 109): `// Caller must lock`.

ColinSullivan1 (changes requested):
Thanks for the contribution!  

The conn lock is needed to properly handle unsubscribe called from application code. 

Good catch with the subscription lock.  That can be removed.

Could you resubmit the PR using only the changes to Subscription.cs?

NATS.Client/Conn.cs
ColinSullivan1: Th connection lock also protects statistics being gathered.

tamnm:
@ColinSullivan1 I reverted changes in conn.cs, added comment into Subscription

ColinSullivan1 (approved):
LGTM - thanks for the contribution!  After tests pass, I'll merge.

ColinSullivan1:
AppVeyor is having issues that seem related to resource constraints.  I tested manually and all tests pass.  Thanks again for the contribution.

PR #195, opened 2017-12-15
Ensure that the connection is disposed of if an exception happens while connecting

Ksero:
I had an application which could end up with hundreds of NATS threads if no server was available. The application would call CreateConnection repeatedly to try to establish a connection.


ColinSullivan1 (commented):
LGTM.  Thanks!  Let me run a few additional stress tests and if all pass, I'll merge.

ColinSullivan1 (approved):
LGTM.  Thanks!

PR #200, opened 2018-01-29
Slow consumer improvements

sebastien-curutchet:
Improve client behavior on big messages.
Avoid to lost publish call if the network connection is closed by the server

### DIFF
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -1538,6 +1538,11 @@ namespace NATS.Client
         {
             bool disconnected = false;
 
+            if (isConnecting() || isClosed() || isReconnecting())
+            {
+                return;
+            }
+
             lock (mu)
             {
                 if (isConnecting() || isClosed() || isReconnecting())
@@ -2250,7 +2255,7 @@ namespace NATS.Client
         // publish is the internal function to publish messages to a nats-server.
         // Sends a protocol data message by queueing into the bufio writer
         // and kicking the flush go routine. These writes should be protected.
-        internal void publish(string subject, string reply, byte[] data, int offset, int count)
+        internal void publish(string subject, string reply, byte[] data, int offset, int count, bool isRecursiveCall = false)
         {
             if (string.IsNullOrWhiteSpace(subject))
             {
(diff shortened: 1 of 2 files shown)


stephenjannin:
We had a production issue where too many big messages where sent during many hours. This had two impacts : 
- The **GNatsd cluster got damaged** (one node was considered as a slow consumer and was excluded from the mesh)
- The C# clients loosed the connectivity to the cluster and could **never reconnect**.
System.NotSupportedException: Stream does not support writing.
   at System.IO.BufferedStream.Write(Byte[] array, Int32 offset, Int32 count)
   at NATS.Client.Connection.publish(String subject, String reply, Byte[] data, Int32 offset, Int32 count)
   at NATS.Client.Connection.Publish(String subject, String reply, Byte[] data)

ColinSullivan1:
@sebastien-curutchet , @stephenjannin , Thank you for this PR, much appreciated!  

Because this change affects reconnect, I have some long running stress tests that I'll run.  Would you have an isolated test to reproduce this? It'd be great to keep one around to test future changes against this problem.

Regarding the NATS server cluster...   It sounds like the cluster was behaving as expected, protecting the cluster as a whole by removing the server that couldn't keep up.  Were clients fairly distributed across server nodes?

sebastien-curutchet:
Hello,

Test Platform :
*Server : 3 nats daemon with a latency from 3 to 9ms (the latency can be generated by an external tool)
     Each daemon has a config file with :
        - routes of the 2 others servers with authorization (timeout: 8)
        - no_advertise=true
        - client authorization activated  (timeout: 8)
        - secure communication activated (tls setup)

* Client : 
    - The client knows the 3 servers (explicit routes in the connection Options)
    - Publish to the subject "foo" a set of files (around 1800 files) from 4KB to 150MB (of course some files are rejected because the payload size is over the maxPayload). Do it in a Parallel.ForEach with a MaxDegreeOfParallelism equals to 200. We did the tests with a MaxDegreeOfParallelism equals to 1 then to 2 before to resolve the Reconnect issue we have, then the SlowConsumer error we have.
    - Subscribe to the subject "foo" using the same connection
    - For each sent files we increment a counter with Interlocked.Increment
    - For each received files we increment another counter with Interlocked.Increment (subscription callback)
    - Just add a sleep of few seconds once all files are sent, to be sure you received all messages
    - Assert the 2 counters are equal.

Our test sends more around 1800 files with a total of more than 1GB in less than 1min30.
If your want to stress the reconnect behavior, just let the lock in the processMsg method. I don't have found yet a good way to slow down the reception only, to force the server to consider the client as a SlowConsumer.

If you use 2 different connections for publication and subscription, the behavior is different and you are not consider as a SlowConsumer by the server, so the Reconnect procedure is not stressed.

NATS.Client/Conn.cs
```diff
@@ -1939,50 +1948,46 @@ internal void processMsg(byte[] msg, long length)
             bool maxReached = false;
             Subscription s;
 
-            lock (mu)
-            {
-                stats.inMsgs++;
-                stats.inBytes += length;
+            Interlocked.Increment(ref stats.inMsgs);
```
ColinSullivan1, line 1951: Thanks, this is something I've been wanting to change.

NATS.Client/Conn.cs
```diff
@@ -2268,39 +2273,52 @@ internal void publish(string subject, string reply, byte[] data, int offset, int
...
+                    stats.outMsgs++;
+                    stats.outBytes += count;
 
-                kickFlusher();
+                    kickFlusher();
+                }
+            }
+            catch (Exception e)
+            {
+                Trace.TraceWarning("Publication failed. Retrying. The error was " + e);
```
ColinSullivan1, line 2312: IMO this trace statement should be removed.

NATS.Client/Conn.cs
```diff
@@ -1534,6 +1538,11 @@ private void processOpError(Exception e)
         {
             bool disconnected = false;
 
+            if (isConnecting() || isClosed() || isReconnecting())
```
ColinSullivan1, line 1541: Were you having locking issues?  I see this same check a few lines below under the lock.

NATS.Client/Conn.cs
```diff
@@ -2268,39 +2273,52 @@ internal void publish(string subject, string reply, byte[] data, int offset, int
...
-                stats.outMsgs++;
-                stats.outBytes += count;
+                    stats.outMsgs++;
+                    stats.outBytes += count;
 
-                kickFlusher();
+                    kickFlusher();
+                }
+            }
+            catch (Exception e)
```
ColinSullivan1, line 2310: The NATSConnectionClosedException is being swallowed here, changing behavior of publishing on a closed connection.

NATS.Client/Conn.cs
ColinSullivan1, line 2312: I think you need:
```C#
if (e is NATSConnectionClosedException)
   throw;
```
(or equivalent) here.

NATS.Client/Conn.cs
sebastien-curutchet: This test is done to avoid a lock to check the state. 
But if 2 threads do the test "at the same time", both return false. So the test have to be done in the lock again to avoid 2 reconnect tasks at the same time.

NATS.Client/Conn.cs
sebastien-curutchet: Yes and I forgot the else states. I will take look at it.

ColinSullivan1:
Closing due to inactivity.  Please don't hesitate to re-open if you'd like to update this with the requested changes.

PR #203, opened 2018-03-09
ToString() fixed name check

devKlausS:
printed null if name was set


ColinSullivan1 (approved):

Issue #210, opened 2018-05-03
NATS client mixes up replies in case of concurrent requests


maksimenko-stanislav:
Hello,

We use NATS and its C# client intensively in our project. Some time ago we started to do load tests and discovered that replies received using IConnection.RequestAsync may not match source requests. I attached an example ([NatsRequestReplyIssue.zip](https://github.com/nats-io/csharp-nats/files/1970586/NatsRequestReplyIssue.zip)) for reproducing such case.

In short, steps to reproduce:

1. Start async subscription using IConnection.SubscribeAsync(string subject, string queue, EventHandler<MsgHandlerEventArgs> handler)
2. **Important note**: handler passed on step 1 is async (see attached example), inside it should be at least one await, i.e. database request, etc. In case of missing async work, all is going fine.
3. Create sample request and response types:
```csharp
        public class Request
        {
            public Guid Id { get; set; }
        }

        public class Reply
        {
            public Guid RequestId { get; set; }
        }
```
4. In message handler assign Request.Id to Reply.RequestId 
5. Start multiple concurrent request.
6. Check Reply.RequestId after deserialization.

NATS versions: 1.0.4, 1.1.0
NATS client: 0.8.0 (latest nuget)

PR #211, opened 2018-05-03
Fix for #210 - race condition in case of concurrent requests

maksimenko-stanislav:
This change fixes issue #210. Root cause: state variable msgHandlerArgs which is updated concurrently. Now new instance of MsgHandlerEventArgs is allocated for each Msg.

### DIFF
--- a/NATS.Client/AsyncSub.cs
+++ b/NATS.Client/AsyncSub.cs
@@ -35,7 +35,6 @@ namespace NATS.Client
         /// </summary>
         public event EventHandler<MsgHandlerEventArgs> MessageHandler;
 
-        private MsgHandlerEventArgs msgHandlerArgs = new MsgHandlerEventArgs();
         private Task                msgFeeder = null;
 
         private bool started = false;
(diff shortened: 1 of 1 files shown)


ColinSullivan1:
@maksimenko-stanislav , thank you for the PR!  I'll take a look.

maksimenko-stanislav:
@ColinSullivan1 , do you have any plans to release new version of csharp NATS client in near future? We need current fix to be included in new release, because at the moment using version 0.8.0 we have to use syncronization mechanisms while sendind requests.

ColinSullivan1:
@maksimenko-stanislav , I understand and will aim to get a release done in the next few weeks.

sixlettervariables:
I've tested this PR locally and found it definitely fixes the problem described in #210. We've seen "dropped" messages in production, but always assumed them to be due to network issues or our own internal race conditions. I have this fix merged against our local setup and have been running it without issue for the last ~4 days.

ColinSullivan1:
@sixlettervariables , thank you for the testing!  Did you see much of a performance impact?  (I plan to run perf tests, but just curious).

sixlettervariables:
@ColinSullivan1 our app handles too little traffic (~20 msg/s) to do real perf testing. That being said, I've noticed no appreciable impact.

maksimenko-stanislav:
@sixlettervariables , @ColinSullivan1 , there is one extra GC allocation in code of PR. I didn't research, but i suppose that it don't leave Gen0 in most cases.

ColinSullivan1:
Consuming about 1.4 M msgs/sec on my small windows VM, so good to go.  Thanks for the contribution!!!

maksimenko-stanislav:
@ColinSullivan1 , thank you very much for fast reaction!

PR #212, opened 2018-05-18
Migrate to VS17 and update dependencies

ColinSullivan1:
Move to a more current build environment and update the .NET core dependencies to current versions.
 
* Change project files to VS 2017 format
* Move to NET 2.0 dependencies
* Fix a bug found closing a connection
* Fix a test failing due to test parameters


NATS.Client/Conn.cs
```diff
@@ -3564,7 +3564,13 @@ private void close(ConnState closeState, bool invokeDelegates)
                             bw.Flush();
                         }
                         catch (Exception) { /* ignore */ }
-                        bw.Dispose();
+
+						try
```
kozlovic: tabs?

kozlovic (approved):
LGTM

NATS.Client/Conn.cs
ColinSullivan1: Thanks Ivan - will fix.

cjbhaines:
@ColinSullivan1 can we put the NATS.Client Library back to .NET Standard 1.6 please? By moving forward to 2.0 we have reduced the comparability and forces all consumers to be on 2.0. The best practice guide is to target the lowest possible version of .NET Standard. I just tried this locally and it still builds ok. This still gains the benefits of the new SDK.

ColinSullivan1:
..True.  I'll move it back.

cjbhaines:
Thanks a lot!

PR #217, opened 2018-06-14
Add the ability to infinitely reconnect

ColinSullivan1:
This PR adds a new `Options.ReconnectForever` value that can be set on `Options.MaxReconnects` to specify that a connection will infinitely attempt to reconnect.


sfrooster:
Great!  When should we expect to see this available in NuGet?

ColinSullivan1:
v0.8.1 with this feature is up on Nuget.  Check it out at:   https://www.nuget.org/packages/NATS.Client/

PR #228, opened 2018-09-06
don't force adding scheme prefix to string

harrisa1:
users should be expected to write reasonable code, this also matches the rest of the clients, i.e. see https://github.com/nats-io/java-nats/blob/master/src/main/java/io/nats/client/Options.java line 584

### DIFF
--- a/NATS.Client/Srv.cs
+++ b/NATS.Client/Srv.cs
@@ -32,12 +32,14 @@ namespace NATS.Client
 
         internal Srv(string urlString)
         {
-            Uri uri;
-            if (!Uri.TryCreate(urlString, UriKind.Absolute, out uri) &&
-                !Uri.TryCreate(defaultScheme + urlString, UriKind.Absolute, out uri)) throw new UriFormatException();
-            var builder = new UriBuilder(uri);
-            builder.Port = builder.Port == noPortSpecified ? defaultPort : builder.Port;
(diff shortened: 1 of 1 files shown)


harrisa1:
for our use case we need to use the tcp scheme, but this forces us to use nats, our other languages don't have this problem

kozlovic:
@harrisa1 Actually all other supported clients are moving to support simplified URLs. That is, if no scheme/port is specified, the default scheme and port are added.

harrisa1:
that sounds great, but this client didn't do that, it forced the string to start with "nats://" even if you might have already supplied a scheme, and didn't touch the port at all, i can update this request with the changes to support defaults, but that is not what this code was doing, if like us you specified a scheme that wasn't "nats://" this prefixed a second scheme which obviously wouldn't work

kozlovic:
@harrisa1 Understood that current code is wrong to prepend nats:// without checking if a scheme is already set. I think that it would be addressed with bringing the changes done for other clients, but I am sure that @ColinSullivan1 would not mind a PR ;-)

harrisa1:
no problem, updated with support for simplified urls and defaults

ColinSullivan1 (approved):
LGTM, I'm running some automated tests, when they pass I'll merge.  Thanks!

PR #232, opened 2018-10-11
Fix race condition under parallel request calls on an encoded connection

### LINKED ISSUE #231: Parallel Requests
I don't understand if this is a bug, or I am just using the client library in a wrong way ... 
The scenario is that I have a service task (connection1 in this example) that reply to request from various source in parallel.
When I have just a normal "for" loop for connection2, all went fine, but If I use Parellel.For strange exception happens and the responses are wrong. I eventually get the message ```(dataIn.QueryIn != dataOut.QueryOut)``` and this is absolutely wrong in my opinion.

The exception i got is ```NATS.Client.NATSException: Unable to deserialize message.``` (from EncodedConn.cs, line 243). This is strange cause the class is perfectly serializable and the error show up only when parallelism is "high" (if 10 parallel requests can be considered "high").

this is the code:

```
using System;
using System.Threading;
using System.Threading.Tasks;
using NATS.Client;


namespace NATSPublisher
{
    class Program
    {
        {
            var random = new Random();
            var connection1 = new ConnectionFactory().CreateEncodedConnection("nats://inner-nats01.test.com:4242");

            connection1.SubscribeAsync("test1", (sender, eventArgs) =>
            {
                // take some time to respond, use a separate task
                Task.Run(() =>
                {
                    try
                    {
                        if (eventArgs.ReceivedObject is LongDataClassResponse dataIn)
                        {
                            Thread.Sleep(1000 + random.Next(100, 1000));
                            connection1.Publish(eventArgs.Reply,
                                new LongDataClassResponse() {QueryIn = dataIn.QueryIn, QueryOut = dataIn.QueryIn});
                        }
                        else
                        {
                            Console.WriteLine("=======================< (null dataIn)");
                        }
                    }
                    catch (Exception ex)
                    {
                        Console.WriteLine("=======================< {0}: {1}", ex.GetType(), ex.Message);
                    }
                });
            });

            var connection2 = new ConnectionFactory().CreateEncodedConnection("nats://inner-nats01.test.com:4242");
            Parallel.For(0, 10, i =>
            {
                try
                {
                    var dataIn = new LongDataClassResponse() {QueryIn = i.ToString()};

                    if (connection2.Request("test1", dataIn, 9000) is LongDataClassResponse dataOut)
                    {
                        Console.WriteLine(i);

                        if (dataOut.QueryIn != dataOut.QueryOut) Console.WriteLine("=======================> (dataOut.QueryIn != dataOut.QueryOut)");

                        if (dataIn.QueryIn != dataOut.QueryOut) Console.WriteLine("=======================> (dataIn.QueryIn != dataOut.QueryOut)");
                    }
                }
                catch (Exception ex)
                {
                    Console.WriteLine("=======================> {0}: {1}", ex.GetType(), ex.Message);
                }
            });

            Console.WriteLine("press any key");
            Console.ReadKey();

            connection1.Close();
            connection2.Close();
        }

        [Serializable]
        public class LongDataClassResponse
        {
            public string QueryIn { get; set; }
            public string QueryOut { get; set; }
        }


    }
}

```

any help very appreciated .... 

Best regards,
Luca

ColinSullivan1:
With changes in the request API, locking changed up a bit so encoded connections default serialization was no longer thread-safe.  This fix:

* Protects memory buffers used for serialization under a lock
* Returns newly allocated byte arrays vs reusing byte arrays when serializing data.

Resolves #231 

### DIFF
--- a/NATS.Client/EncodedConn.cs
+++ b/NATS.Client/EncodedConn.cs
@@ -95,7 +95,7 @@ namespace NATS.Client
     public class EncodedConnection : Connection, IEncodedConnection
     {
         private MemoryStream sStream   = new MemoryStream();
-        byte[] sBytes = new byte[1024];
+        private object       sStreamLock = new object();
 
         private MemoryStream dStream     = new MemoryStream();
         private object       dStreamLock = new object();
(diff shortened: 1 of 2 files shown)


kozlovic (approved):
LGTM

ColinSullivan1:
Thanks @kozlovic.

PR #234, opened 2018-10-12
Fix a memory leak in request/reply

### LINKED ISSUE #227: Request-reply holding onto reference
Seems the reply message isn't freed when using a long timeout on `IConnection.Request`:

```
using System;
using System.Threading;
using NATS.Client;

namespace RefLeak
{
    class Program
    {

        static void Main()
        {
            using (var connection = new ConnectionFactory().CreateConnection("nats://localhost:4222"))
            {
                var subject = "subject";
                connection.SubscribeAsync(subject, OnReceive);

                while (true)
                {
                    var data = new byte[100000];
                    var reply = connection.Request(subject, data, int.MaxValue);
                    Thread.Sleep(500);
                    GC.Collect();
                    Console.WriteLine($"Total memory: {GC.GetTotalMemory(false)}");
                }
            }
        }

        private static void OnReceive(object sender, MsgHandlerEventArgs args)
        {
            var data = new byte[100000];
            ((IAsyncSubscription)sender).Connection.Publish(args.Message.Reply, data);
        }
    }
}
```

Sample output:

```
Total memory: 703416
Total memory: 806968
Total memory: 911080
Total memory: 1009144
Total memory: 1113304
Total memory: 1211320
Total memory: 1315528
```

This is with Core 2.1.

Edit:

I think this comes from the CancellationTokenSource allocated in the InFlightRequest ctor not being disposed (which would dispose the timer callback that ultimately holds onto the reply message).

ColinSullivan1:
Various resources associated with the timer were being leaked.   This fix tracks and disposes of the resources by disposing the token cancellation registration for the timeout cancellation token, and the primary token for the in-flight request. 

Resolves #227 

### DIFF
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -148,7 +148,7 @@ namespace NATS.Client
             new Dictionary<string, InFlightRequest>(StringComparer.OrdinalIgnoreCase);
 
         // Handles in-flight requests when using the new-style request/reply behavior
-        private sealed class InFlightRequest
+        private sealed class InFlightRequest : IDisposable
         {
             public InFlightRequest(CancellationToken token, int timeout)
             {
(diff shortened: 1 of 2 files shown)


kozlovic (approved):
LGTM

ColinSullivan1:
Thanks @kozlovic ...

PR #242, opened 2018-12-17
Default the count of published messages to 1

kiates:
It seems that 1 would be a better default count for the publish example.


watfordgnf (approved):
Looks good, thanks!

PR #245, opened 2019-02-16
Couple of small fixes

lanfeust69:
See messages of individual commits

### DIFF
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -86,7 +86,7 @@ namespace NATS.Client
 
         // NOTE: We aren't using Mutex here to support enterprises using
         // .NET 4.0.
-        readonly internal object mu = new Object();
+        readonly private object mu = new Object();
 
         private Random r = null;
 
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -414,7 +414,8 @@ namespace NATS.Client
 
                     client = new TcpClient();
                     var task = client.ConnectAsync(s.url.Host, s.url.Port);
-                    task.ContinueWith(t => t.Exception);
+                    // avoid raising TaskScheduler.UnobservedTaskException if the timeout occurs first
+                    task.ContinueWith(t => GC.KeepAlive(t.Exception), TaskContinuationOptions.OnlyOnFaulted);
                     if (!task.Wait(TimeSpan.FromMilliseconds(timeoutMillis)))
                     {
                         client = null;
@@ -2486,7 +2487,8 @@ namespace NATS.Client
         private InFlightRequest setupRequest(int timeout, CancellationToken token)
         {
             InFlightRequest request = new InFlightRequest(token, timeout);
-            request.Waiter.Task.ContinueWith(t => t.Exception);
+            // avoid raising TaskScheduler.UnobservedTaskException if the timeout occurs first
+            request.Waiter.Task.ContinueWith(t => GC.KeepAlive(t.Exception), TaskContinuationOptions.OnlyOnFaulted);
             bool createSub = false;
             lock (mu)
             {
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -413,7 +413,8 @@ namespace NATS.Client
                     }
 
                     client = new TcpClient();
-                    var task = client.ConnectAsync(s.url.Host, s.url.Port).ContinueWith(t => t.Exception);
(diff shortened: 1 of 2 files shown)


NATS.Client/Conn.cs
```diff
@@ -413,7 +413,8 @@ internal void open(Srv s, int timeoutMillis)
                     }
 
                     client = new TcpClient();
-                    var task = client.ConnectAsync(s.url.Host, s.url.Port).ContinueWith(t => t.Exception);
+                    var task = client.ConnectAsync(s.url.Host, s.url.Port);
+                    task.ContinueWith(t => t.Exception);
```
sixlettervariables:
```suggestion
                    task.ContinueWith(t => GC.KeepAlive(t.Exception), TaskContinuationOptions.OnlyOnFaulted);
```

NATS.Client/EncodedConn.cs
```diff
@@ -166,15 +166,12 @@ internal object defaultDeserializer(byte[] data)
 
         private void publishObject(string subject, string reply, object o)
         {
-            lock (mu)
-            {
-                if (onSerialize == null)
-                    throw new NATSException("IEncodedConnection.OnSerialize must be set (.NET core only).");
+            if (onSerialize == null)
```
sixlettervariables, line 169: Should protect against `onSerialize` changing between the null check and the call:
```csharp
    Serializer serializer = onSerialize;
    if (serializer == null)
        throw new NATSException("IEncodedConnection.OnSerialize must be set (.NET core only).");

    byte[] data = serializer(o);
```

sixlettervariables (changes requested):
Couple of changes.

NATS.Client/Conn.cs
lanfeust69: I don't understand the purpose of `GC.KeepAlive` here.

Regarding limiting the continuation to Faulted tasks, it probably makes sense, but should be done separately (along with the `GC.KeepAlive` if deemed useful) to address the other places where this pattern has been introduced in dde17f208afeb9133bf45b23fa4683b26cd7ad49 and 81f08b0d2d34dfc4af66fa7dffbfeaac442f8820.

NATS.Client/EncodedConn.cs
lanfeust69: Is it really a good idea ?

In practice, `OnSerialize` will be set once and never touched, and we only want a better error than the dreaded `NullReferenceException` if not set.

OTOH, if by extraordinary it is a voluntary change, maybe using the "old" one after the change could be a real problem (for instance there could be some kind of disposable buffer involved, which has been disposed). Seems worse than changing the type of the exception, I think. And of course, adding a proper lock mechanism is surely too expensive for such an unlikely scenario.

NATS.Client/Conn.cs
watfordgnf: It ensures the unused exception is truly observed, and not thrown away. Good catch on the other areas which could use that.

lanfeust69: Sorry, still don't get it...

According to the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.gc.keepalive), `GC.KeepAlive` :

> References the specified object, which makes it ineligible for garbage collection from the start of the current routine to the point where this method is called.

But it is already referenced, as it is about to be set as the `Result` of the `Task<Exception>` created by `ContinueWith()`, isn't it ?

watfordgnf: This is a pattern seen in https://github.com/StackExchange/StackExchange.Redis/blob/master/src/StackExchange.Redis/TaskExtensions.cs

I've reached out to @mgravell for more details.

lanfeust69: I get it now : the `GC.KeepAlive()` isn't useful for "observing" the Exception (that was already OK), but to change the `ContinueWith()` overload that will be used : it is now the one that takes an `Action<Task>`, and **returns a simple `Task`** (instead of a `Task<Exception>` for the current overload that accepts a `Func<Task, Exception>`). As this task will always complete synchronously, it might be optimized to return the singleton `Task.CompletedTask` (when available), avoiding an allocation.

So this is actually a small optimization, especially when associated with the `OnlyOnFaulted` which should mean few actual continuations. Moreover, this is in the connection step, so probably few calls, it would be more interesting in the other place where this pattern is used (`Connection.setupRequest`).

watfordgnf (approved):

ColinSullivan1 (approved):
LGTM.  Thanks for the contribution!

PR #278, opened 2019-08-23
Fix memory leak on timed out connection attempt

### LINKED ISSUE #236: Memory leak on failed connection attempts.
In operation I would like to be able to leave applications running prior to starting the NATS server that expect to see one at some point in their life.  Currently I have a loop:

```
while ((natsConnection == null) || (natsConnection.State != ConnState.CONNECTED))
{
    try
    {
        Log.Information("Creating nats connection with: [{0}]...", opts.Url);

        natsConnection = new ConnectionFactory().CreateConnection(opts);
    }
    catch (Exception e)
    {
        Log.Information(e.ToString());
    }
}
```

natsConnection remains null on failure but the memory usage steadily grows with each attempt.  the memory is not returned on a successful connection.

ColinSullivan1:
Fix a memory leak upon a failed connection.

Note, upon a first failed connection there will be a bump in memory of a few Kb as the .NET runtime allocates objects required to create a socket.  This is a one time allocation.  With this fix, memory remains stable after that point.

In debugging this, some tasks associated with the TCP client connection were collected while others were not, so there was not a one-to-one leaked object per failed attempt.  No leaked objects had references to, or were referenced by NATS.Client objects - at least that the memory profiler would report.  I did not get to the bottom of that mystery, but quantitatively testing a loop of 1000 failed connection attempts measuring memory + object counts indicated this bug was fixed with the changes.

Resolves #236.

### DIFF
--- a/NATS.Client/Conn.cs
+++ b/NATS.Client/Conn.cs
@@ -439,6 +439,7 @@ namespace NATS.Client
                     task.ContinueWith(t => GC.KeepAlive(t.Exception), TaskContinuationOptions.OnlyOnFaulted);
                     if (!task.Wait(TimeSpan.FromMilliseconds(timeoutMillis)))
                     {
+                        close(client);
                         client = null;
                         throw new NATSConnectionException("timeout");
                     }
@@ -467,7 +468,7 @@ namespace NATS.Client
                 return false;
             }
 
-            internal void closeClient(TcpClient c)
+            internal static void close(TcpClient c)
             {
                 if (c != null)
                 {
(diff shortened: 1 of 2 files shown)


watfordgnf:
We may not dispose of `sslStream` if there is an exception during `makeTls`:

https://github.com/nats-io/nats.net/blob/d9993456439c1704ca819033eac155308db2855c/NATS.Client/Conn.cs#L483-L507

ColinSullivan1:
Good catch, thanks @watfordgnf. I'll add that to this PR.

watfordgnf (approved):
LGTM pending CI.

ColinSullivan1:
@watfordgnf, added one more small thing - nulled out the ssl stream after dispose for good measure.  I think this should do it.  Tests are green locally, will let you know when they're green for CI.

ColinSullivan1:
CI Is green.

Issue #281, opened 2019-08-26
Add Message.Respond()


ColinSullivan1:
See:
- https://github.com/nats-io/nats.go/pull/472
- https://github.com/nats-io/nats.go/pull/489

ColinSullivan1:
Resolved by #283

PR #283, opened 2019-08-27
Add Message.Respond implementing #281

watfordgnf:
Implements #281 by adding Respond to Msg to simplify responses to Request-Reply.

I brought over the same changes to Subscription that landed in https://github.com/nats-io/nats.go/pull/489/files however this breaks TestConnection.TestClosedConnections which had certain assumptions about the exception messages.

Should we alter the unit test to keep parity with the go client, or should we alter the code to keep raising the same exception?

### DIFF
--- a/NATSUnitTests/UnitTestSub.cs
+++ b/NATSUnitTests/UnitTestSub.cs
@@ -1046,28 +1046,28 @@ namespace NATSUnitTests
             using (new NATSServer())
             {
                 using (IConnection c = utils.DefaultTestConnection)
+                using (ISyncSubscription s = c.SubscribeSync("foo"))
                 {
-                    ISyncSubscription s = c.SubscribeSync("foo");
-
                     string replyTo = c.NewInbox();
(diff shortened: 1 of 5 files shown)


NATSUnitTests/UnitTestSub.cs
```diff
@@ -1040,6 +1039,70 @@ public void TestInvalidSubjects()
                 }
             }
         }
+
+        [Fact]
+        public void TestRespond()
+        {
+            using (new NATSServer())
```
ColinSullivan1, line 1046: Would you mind adding a test that calls respond after the connection has been closed gracefully, and then also after the server has been closed and reconnects have failed (or just set Options.AllowReconnect == false)?

ColinSullivan1:
As much as I want to keep parity, imo we shouldn't change the exception being thrown.  While unlikely users would run into that, it would break backward compatibility.  We should keep the current exception...

ColinSullivan1:
Thanks!  LGTM.

ColinSullivan1 (approved):

PR #286, opened 2019-09-23
Fixed various typos in comments and field names. Fixed Exception for passing comment correctly

CycoPH:
- Fixed various typos in comments
- Fixed the naming of a field. ArrivalSubscription had a typo
- Fixed bug in NATSBadSubscriptionException.
- The passed error message was never forwarded to the NATSException container

### DIFF
--- a/NATS.Client/Msg.cs
+++ b/NATS.Client/Msg.cs
@@ -159,10 +159,19 @@ namespace NATS.Client
             this.data = data;
         }
 
-        /// <summary>
-        /// Gets the <see cref="ISubscription"/> which received the message.
-        /// </summary>
-        public ISubscription ArrivalSubscription
+		/// <summary>
+		/// Gets the <see cref="ISubscription"/> which received the message.
+		/// </summary>
+		[ObsoleteAttribute("This property will soon be deprecated. Use ArrivalSubscription instead.")]
+		public ISubscription ArrivalSubcription
+		{
+			get { return sub; }
+		}
+
+		/// <summary>
+		/// Gets the <see cref="ISubscription"/> which received the message.
+		/// </summary>
+		public ISubscription ArrivalSubscription
         {
             get { return sub; }
         }
--- a/NATS.Client/Msg.cs
+++ b/NATS.Client/Msg.cs
@@ -162,7 +162,7 @@ namespace NATS.Client
         /// <summary>
         /// Gets the <see cref="ISubscription"/> which received the message.
         /// </summary>
-        public ISubscription ArrivalSubcription
+        public ISubscription ArrivalSubscription
         {
             get { return sub; }
         }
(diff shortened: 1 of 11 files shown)


NATS.Client/Msg.cs
```diff
@@ -162,7 +162,7 @@ public void AssignData(byte[] data)
         /// <summary>
         /// Gets the <see cref="ISubscription"/> which received the message.
         /// </summary>
-        public ISubscription ArrivalSubcription
+        public ISubscription ArrivalSubscription
```
watfordgnf, line 174: This is a breaking change to the API (but a very reasonable one). Perhaps we should consider the original misnamed property forwarding to the new property, with the original property marked as Obsolete?

@ColinSullivan1

watfordgnf (commented):
I have no issues with the changes at hand, just we need to consider the implications of `Msg.ArrivalSubscription` being renamed.

NATS.Client/Msg.cs
ColinSullivan1: Yes, I agree.  Wow, am I embarrassed at that one.   Yep, we should keep both, and mark `ArrivalSubcription` obsolete.  We can then remove it for version 1.0.

ColinSullivan1:
LGTM, other than the comments above.  Thank you for the contribution!

NATS.Client/Msg.cs
ColinSullivan1: @cycoPH, would you mind adding a new property (correcting my spelling), adding an obsolete attribute, e.g.

```[ObsoleteAttribute("This property will soon be deprecated. Use ArrivalSubscription instead.")]```

...and have the old property call the new one?

Comment (CycoPH) [reply NATS.Client/Msg.cs:174]: Sure did the change a resubmitted

File: NATS.Client/Msg.cs [hunk 2]
```diff
@@ -159,19 +159,19 @@ public void AssignData(byte[] data)
...
-
-		/// <summary>
-		/// Gets the <see cref="ISubscription"/> which received the message.
-		/// </summary>
-		public ISubscription ArrivalSubscription
+        /// <summary>
+        /// Gets the <see cref="ISubscription"/> which received the message.
+        /// </summary>
+        [ObsoleteAttribute("This property will soon be deprecated. Use ArrivalSubscription instead.")]
+        public ISubscription ArrivalSubcription
```
watfordgnf, line 166: Should be able to make this a one liner:
```csharp
public ISubscription ArrivalSubcription => ArrivalSubscription;
```

ColinSullivan1:
@CycoPH , we've incorporated your changes in #346.  There were so many changes since this this PR was created it was much easier to adopt your changes in a new one.  You've been mentioned in the new PR and will be in the release notes.  We really appreciate your contribution!

PR #288, opened 2019-09-27
Consolidates solution/projects and moves to Azure DevOps

danielwertheim:
Main changes are:
  
  - Using .NET Core SDK style projects to multitarget
    .NET4.5 and .NET Standard 1.6
  - Organizes folders and projects
  - Adds Directory.builds.props to unify props in all projects
  - Splits tests into IntegrationTests and UnitTests
  - Moves signing key to root
  - Fixes pointing to correct logo at CNCF artwork in NuGet
  - Uses Azure Pipeline (YAML) for CI
  - Starts using TestCollection to control paralellization
  - Uses Global.json to lock to certain SDK
    (be as similar as possible at build server)

PR #289, opened 2019-10-01
Fixes so that SemVer is not pre-release when build is against master

danielwertheim:
When the PR #288 was merged into `master` yesterday, I noticed the pipeline generated the wrong `SemVer` as it was counted as a pre-release. The idea is that against `master` it is seen as a stable release.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -4,9 +4,10 @@ variables:
   Ver: '0.9.0'
   BuildRev: $[counter(variables.Ver, 1)]
   BuildVer: $[ format('{0}.{1}', variables.Ver, variables.BuildRev) ]
-  SemVer: $[ format('{0}-pre{1}', variables.Ver, variables.BuildRev) ]
-  ${{ if eq( variables.Build.SourceBranchName, 'master' ) }}:
-    SemVer: ${{$(Ver)}}
+  ${{ if eq( variables['Build.SourceBranchName'], 'master' ) }}:
+    SemVer: $[ variables.Ver ]
+  ${{ if ne( variables['Build.SourceBranchName'], 'master' ) }}:
+    SemVer: $[ format('{0}-pre{1}', variables.Ver, variables.BuildRev) ]
 
 trigger:
 - master


danielwertheim:
@ColinSullivan1 @watfordgnf The build is now failing due to the reason that this PR is from my fork and the trigger is enabled to build on PRs from forks. How-ever, when we download the ZIP for the latest NATS-Server from GitHub, we use an token. This token is by default not "shared" with fork PR builds.

image

More info here: https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#validate-contributions-from-forks

Suggestions? Should we take the simplest path by enable sharing tokens or stop building PRs automatically from forks? Apparently we could enable that they are only built if members of this repo comments the PR `/AzurePipelines run` or `/azp run` More info: https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#comment-triggers

watfordgnf:
I think we can safely run them on forks provided:
1. We do not make secrets available (the default).
2. We only use Microsoft hosted agents (which I believe is the current case).

On a personal project I've had success with allowing builds from forks.

watfordgnf (approved):

danielwertheim:
@watfordgnf If we don't enable sharing secrets to forks we need to come up with something for the downloading of Nats Server. We have talked about replacing this with Docker. But we aren't there yet. We do use the MS Hosted Agents.

Given this, are you for sharing the secret or should we not build forks? Or share and build forks but require comment from team members for others PRs?

ColinSullivan1 (approved):
LGTM.

ColinSullivan1:
Stepping back, do you know why a token is required to download to download the NATS server release?  It's in a public location, so shouldn't require a token.

watfordgnf:
The `DownloadGitHubRelease` task requires a service connection, that changes the calculus a bit.

danielwertheim:
The task requires it as far as I understood it.   Can perhaps fork it and tweak it.

danielwertheim:
Will add a simple solution to get it working without that task for now.

PR #290, opened 2019-10-01
Uses native PowerShell and the GitHub API to get the release

danielwertheim:


danielwertheim:
@ColinSullivan1 @watfordgnf  Sorry for all the commits 😊 Anyways, this PR fixes the issue encountered in PR #289 where a fork PR failed due to the download task needed a service connection. I've replaced this with some simple PowerShell script that uses the GitHub API to get the `latest` Windows 64 asset.

watfordgnf (approved):
Looks good!

We may consider having a second pipeline to run to confirm NATS Server 1.X support is maintained. As an example, my shop won't be off 1.X for quite some time given our long install periods (up to 24 months without being touched) and the low likelihood of updating the NATS server binary during a quarterly or biannual update.

danielwertheim:
@watfordgnf  Yes. We could also have two jobs running in parallel testing different versions. I do believe the free-hosted version includes parallel jobs.

But should we have that for all builds or e.g. just against `master`?

watfordgnf:
We could use a separate `yml` file to describe the job, and typically run it for `master`. And then we could also run it OOB if we're working on something that we think may regress on the 1.X line.

That being said, I don't think we'll burn up the free minutes even if we ran both jobs always.

ColinSullivan1:
There are 2.0 tests that wouldn't pass for 1.x, and then newer client features coming that won't work with 2.0.0.  I'd be ok with a simple sanity check test where a 1.x server is used and there is a compatibility test that connects, subscribes and publishes.  Could download the 1.x image and relabel the exe to do it in the same run....  wdyt?

ColinSullivan1:
LGTM.

PR #292, opened 2019-10-03
Refactors the tests in order to be able to specify ports

danielwertheim:
- Introduces the concept of TestSuites and using fixtures Context.
  - Extracts repeated config for tests.
  - Uses Factory methods to unify behavior of creating NATS test server.
  - Assignes custom port ranges to different test classes.
  - Extracts pure unit tests from integration tests.
  - Tries to make test server shutdown a bit more graceful.
  - Lowering the max cycles for the memleak tests.
  - Removes code not being used in tests, e.g. in TLS tests.
  - Includes all mem check tests within the dedicated compiler directive
  - Moves global.json to correct place (Using latest SDK 3.0.100 as that
    is what has been used on build server.)
  - Updates to latest test dependency (NuGet dep).


src/Tests/IntegrationTests/TestConnection.cs
```diff
@@ -1093,12 +961,147 @@ public void TestDrainStateBehavior()
...
 
-        /// NOT IMPLEMENTED:
-        /// TestServerSecureConnections
-        /// TestErrOnConnectAndDeadlock
-        /// TestErrOnMaxPayloadLimit
+    public class TestConnectionMemoryLeaks : TestSuite<ConnectionMemoryLeaksSuiteContext>
+    {
+        public TestConnectionMemoryLeaks(ConnectionMemoryLeaksSuiteContext context) : base(context) { }
+
+#if memcheck
```
danielwertheim, line 970: @ColinSullivan1 @watfordgnf Two tests were lying outside this directive, but I supposed they were meant to be inside that as well or should they always run?

danielwertheim:
Was thinking of changing one more thing but want to hear your thoughts.

Renaming test classes, e.g.

```
TestAuthorization -> AuthorizationTests
TestConnection -> ConnectionTests
TestConnectionMemoryLeaks -> ConnectionMemoryLeakTests
```

I do believe the names `TestConnection` sounds like a "thing", not something that tests the `Connection`.

Also, I do believe the suites/classes will "read better" e.g.

```csharp
public class TestConnection : TestSuite<ConnectionSuiteContext> {}
```

becomes

```csharp
public class ConnectionTests: TestSuite<ConnectionTestsContext> {}
```

watfordgnf:
I'm all for the renaming if it makes sense. Nobody should have a dependency on these test names anyways.

watfordgnf (approved):
Definitely enjoy these changes and am all for renaming to improve the 'readability' (so to speak).

src/Tests/IntegrationTests/TestConnection.cs
ColinSullivan1: They were intended to be run, but I had to comment them out to occasionally run locally...  In the AppVeyor free tier, I was occasionally getting resource related issues that didn't reflect what was going on locally as reported by the monitoring tools.  I suspect Pipelines will be much better here - let's give it a go.  Adding them doesn't need to hold up this PR though - can be done in another.

ColinSullivan1:
This renaming is great, and like @watfordgnf mentioned, no-one should have dependencies on them - at least production related.

ColinSullivan1 (approved):
LGTM.

PR #294, opened 2019-10-08
Deploy on tag

danielwertheim:
Changes so that:
- Pull requests against `master` and direct commits to `master` will: build, test and create a package but **will not push** any package to any feed. It will make it available via Azure DevOps portal as an artifact attached to the build. Will use a version that is `yyyy.MM.dd-pre{datecounter}` e.g `2019.10.01-pre1`
- Tags: Will use the annotated name of the tag as the version for the package and will build, test, create the package **as well as push the package to NuGet.Org**.

When creating a release, **do not use a** `v`prefix as the tag version as that is used as the version for the binaries, packages etc.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -135,9 +135,10 @@ stages:
           steps:
           - checkout: none
           - task: NuGetCommand@2
-            displayName: 'Push Nupkg to AZ Feed'
+            displayName: 'Push Nupkg to NuGet.Org'
             inputs:
               command: push
-              packagesToPush: '$(Pipeline.Workspace)/**/*.nupkg;!$(Pipeline.Workspace)/**/*.symbols.nupkg'
-              publishVstsFeed: 'lab-feed'
+              nugetFeedType: external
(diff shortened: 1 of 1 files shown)


watfordgnf (approved):

azure-pipelines.yml
```diff
@@ -1,119 +1,144 @@
...
-
-        Write-Host "Downloading asset $name to $outputFile from $assetUrl"
-        New-Item -ItemType Directory -Force -Path $(Agent.TempDirectory)\nats-server
-        (New-Object System.Net.WebClient).DownloadFile($assetUrl, $outputFile)
+stages:
+- stage: Build
+  jobs:
+  - job: BuildArtifacts
+    displayName: 'Builds, tests & produces artifacts'
+    timeoutInMinutes: 10
```
ColinSullivan1: Can you increase this to 20?  As we add tests or if resources are constrained, we might need the extra time.

danielwertheim: Done

ColinSullivan1 (approved):
LGTM.

PR #295, opened 2019-10-09
Changes test class names

danielwertheim:
This was something we discussed in PR #292 Changes the test names to read better. Also updates async tests to return Tasks and adds some assert info to some flaky tests.

After the rename, I actually started to get some flaky tests reported in the CI pipeline. But sinze they all had the same prefix `Test` before, the order or anything should not have changed. And the order is anyhow now not that important since we run them in parallel against different ports.

Coincidence? Opinions?

### DIFF
--- a/src/Tests/IntegrationTests/AuthorizationTests.cs
+++ b/src/Tests/IntegrationTests/AuthorizationTests.cs
@@ -24,9 +24,9 @@ namespace IntegrationTests
     /// <summary>
     /// Run these tests with the gnatsd auth.conf configuration file.
     /// </summary>
-    public class TestAuthorization : TestSuite<AuthorizationSuiteContext>
+    public class AuthorizationTests : TestSuite<AuthorizationTestsContext>
     {
-        public TestAuthorization(AuthorizationSuiteContext context) : base(context) { }
+        public AuthorizationTests(AuthorizationTestsContext context) : base(context) { }
 
         int hitDisconnect;
 
--- a/src/Tests/IntegrationTests/BasicTests.cs
+++ b/src/Tests/IntegrationTests/BasicTests.cs
@@ -36,18 +36,19 @@ namespace IntegrationTests
         {
             using (NATSServer.CreateFastAndVerify())
             {
-                IConnection c = Context.OpenConnection();
+                using (var c = Context.OpenConnection())
+                {
+                    string u = c.ConnectedUrl;
 
-                string u = c.ConnectedUrl;
+                    Assert.False(string.IsNullOrWhiteSpace(u), string.Format("Invalid connected url {0}.", u));
 
-                Assert.False(string.IsNullOrWhiteSpace(u), string.Format("Invalid connected url {0}.", u));
+                    Assert.Equal(Defaults.Url, u);
 
-                Assert.Equal(Defaults.Url, u);
+                    c.Close();
+                    u = c.ConnectedUrl;
 
-                c.Close();
-                u = c.ConnectedUrl;
-
-                Assert.Null(u);
+                    Assert.Null(u);
+                }
             }
         }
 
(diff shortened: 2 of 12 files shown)


danielwertheim:
A bit frustrating. Same behavior confirmed in our pipeline. Failing on `TestRequestAsyncTimeout_OldRequestStyle`. Takes to long suddenly.

@ColinSullivan1  Not sure we should spend time on this now as we talked a bit about removing that style? So we could reject this and redo/integrate the simple work when that has been fixed.

watfordgnf:
I think by this point there would be few remaining folks tied to the old request/reply style (especially those on NATS Server 2.X). It introduces a lot of unnecessary work behind the scenes.

danielwertheim:
If I understood it correctly it should apparently be left in (Slack conv). But it's a bit weird that renaming causes these tests to start failing `Unexpected timeout behavior. Expected lt 5000ms got 15753ms`. I mean, that is not just a "skew".

watfordgnf:
No that seems like unexpected test dependencies (or at least behavior that blocks other tests).

ColinSullivan1:
After discussion with other NATS maintainers, there are use cases that the old style can still apply to (involving) the circuit breaker pattern, and it's used in other ongoing development so let's keep the old style.

ColinSullivan1:
I suppose it's possible renaming might have changed the order, depending on how xunit decides to run the tests.  But with the port changes it shouldn't change things.  Regardless, the timeout should be around 5000 ms when it doesn't receive a message back (which would be an error in the test).  I'd like to track this down, I can help take a look today.

ColinSullivan1:
Even re-running only those tests fail.  This is nuts.

danielwertheim:
But renaming should not really affect it. As we do run them in parallel now and the suffix was the same on all test classes `Test` so there should be no difference in order. It must be something else. Just went through all netcore2.2 tests in `BasicTests`. Ensuring disposal and awaiting tasks etc. in the tests.

jasper-d:
@danielwertheim We have had similar issues with flaky integration tests, though not only with request/reply but also ordinary pub sub where the subscriber never received a message. I never had the time/pressure to investigate them. I *solved* it by excluding them in our automated builds. 😳
We fortunately never observed these problems outside of tests.

A few notes about our environment:

- Our tests use random subject prefixes (i.e. `Guid.NewGuid().ToString("N")`) to ensure there are no "collisions" during tests. 
- All network traffic goes through loopback interface only, so latency shouldn't be an issue. 
- gnatsd logs (log level debug) show nothing unusual.
- I know that we have this problem running on netcore 2.1 and 2.2 (not that I know of any runtime version where he hadn't had that issue), possibly even with .NET 4.x (would need to check git history if we used nats back then).
- Tests are run on Windows boxes 🤨

Cutting a long story short:

- The problem isn't new (for us)
- It is not limited to request/reply (for us)
- It occurs with gnatsd 1.2, 1.4 and 2.0 (I think we never deployed 1.3 and we definitely haven't deployed 2.1 yet)
- It occurs with nats.net 0.9 and 0.8.1

ColinSullivan1:
@jasper-d, thank you for the information, much appreciated.  With @danielwertheim's port changes there shouldn't be any subject name collisions. Moving to loopback in some tests cleared up some tests in our previous CI (AppVeyor).  I rarely see any flapping tests running locally, even on a smaller VM.  I do share your frustration - we'll dig into these.  Just curious, which CI system do you use?

jasper-d:
@ColinSullivan1  Dont worry, there is no frustration on my side, I‘m pretty happy with NATS 😊 

We are currently using Azure Devops Server (on premise) and were previously on TFS Server 2018 (that‘s the predecessor of Devops Server). Test failures reproduce on local dev machines quite well.

danielwertheim:
Closing this one for now and will make a new effort ones all stabilizing changes are in.

PR #296, opened 2019-10-09
Removes unnecessary sync over async for Drain and the lock

danielwertheim:
The entrypoint `DrainAsync` code calls into sync code via a new task, so the sync entrypoint `Drain` should be able to not go via the `Async` member. That way we reduce a sync-over-async usage and `Wait` call. I know the risk has been removed since the async member creates it's own task so deadlocks should not be an issue. How-ever, I do believe creating Tasks and introducing blocking Waits should be kept to a minimum.

Also removes the `lock` for switching status as that is done in the inner `drain` member.

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -3929,6 +3929,11 @@ namespace NATS.Client
             if (timeout <= 0)
                 throw new ArgumentOutOfRangeException(nameof(timeout), "Timeout must be greater than zero.");
 
+            lock (mu)
+            {
+                status = ConnState.DRAINING_SUBS;
+            }
+
             drain(timeout);
         }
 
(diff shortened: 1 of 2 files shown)


danielwertheim:
One test failed so I reverted back the `lock(mu)` and also improved the test a bit. We could remove the `lock(mu)` part and rewrite the test a bit.

```csharp
...
var drainTask = c.DrainAsync(10000);
SpinWait.Unitl(c.IsDraining(), 250);
Assert.True(c.State == ConnState.DRAINING_SUBS);
...
```

watfordgnf (approved):
I'm all for improving the sync vs async plumbing.

watfordgnf:
I wonder how many users would be testing for `IsDraining` after calling `DrainAsync`?

danielwertheim:
Do not know. My knowledge is still to limited. Since I now "fixed" so that the test awaits, I added it as it was asserted "above".

ColinSullivan1 (approved):
LGTM.  Thanks!

PR #301, opened 2019-10-10
Fixes bug with possible null task

danielwertheim:
This member is called from InternalDrain which in turn returns the
resulting task which e.g. is used in Task.WaitAll which will fail
if array of task contains a null element.


watfordgnf (approved):

ColinSullivan1:
LGTM.  Thanks!

PR #302, opened 2019-10-10
Add Connection.FlushBuffer() and flush buffers after requests

### LINKED ISSUE #298: Soft flush
.NET network stacks seem be behave differently on different machines - sometimes data is flushed immediately, other times there's a delay.  We need a "soft" flush API to flush the underlying socket buffer without sending a PING in various cases.

This could be a modification of the `Connection.Flush(timeout)` API to accept a `-1`, defined by a constant, e.g. `Soft`?) that will flush the socket buffer but not send a PING.  We should investigate the performance impact of flushing after every publish.  If is too detrimental to performance, then a soft flush should be called after every latency sensitive API, namely the `Connection.Request` APIs and in `msg.Respond`.

See #297.

ColinSullivan1:
Resolves #298 

### DIFF
--- a/src/Tests/IntegrationTests/TestConnection.cs
+++ b/src/Tests/IntegrationTests/TestConnection.cs
@@ -916,7 +916,7 @@ namespace IntegrationTests
                     c.Publish("foo", null);
                 }
                 Stopwatch sw = Stopwatch.StartNew();
-                Assert.Throws<NATSTimeoutException>(()=> s.Drain(500));
+                Assert.Throws<NATSTimeoutException>(() => s.Drain(500));
                 sw.Stop();
 
                 // add slack for slow CI.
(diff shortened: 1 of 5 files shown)


src/Tests/IntegrationTests/TestConnection.cs
```diff
@@ -967,6 +967,53 @@ public async Task TestDrainStateBehavior()
...
+                opts.ClosedEventHandler = (obj, args) =>
+                {
+                    closed.Set();
+                };
+                opts.DisconnectedEventHandler = (obj, args) =>
+                {
+                    disconnected.Set();
+                };
+
+                c = Context.ConnectionFactory.CreateConnection(opts);
```
danielwertheim: I have noticed that we don't dispose the connection in various places in tests. Is there a reason for not doing this (like, knowing it has unnecessary time consuming cleanup)? I do believe it should be in to cause similar usage as users would have.

danielwertheim (approved):

src/Tests/IntegrationTests/TestConnection.cs
ColinSullivan1: Other than laziness, no...  ;)   Great point - it's good practice to do so.  I'll update this test.

Issue #303, opened 2019-10-15
Add support for IObservable in the existing client


danielwertheim:
Implement support for making `IAsyncSubscription` observable (`IObservable<Msg>`).

Expose it as `INATSObservable<>` under namespace `Rx` so that official `System.Reactive` lib can be brought in and used if user wants to.

The `NATS.Rx` namespace is used to contain implementations and two extension methods for returning the `IObservable<>` implementation:
- `connection.Observe(subject)`
- `asyncSub.ToObservable()`

The `NATS.Rx.Ops` namespace is used (so that you instead can use `System.Reactive` if you want) to provide simple extension methods for allowing:
- Subscribe
- SubscribeSafe
- Where
- Select

**Delegating observer**
Provide extension methods (`Subscribe` and `SubscribeSafe`) for subscribing an observer without passing an observer implementation, but just anonymous methods (delegating observer).

**Subscribe vs SubscribeSafe**
If an observer is misbehaving, `Subscribe` will remove the observer subscription and dispose it so no further dispatches from the observable will be provided to that observer. The `SubscribeSafe` member will keep the observer and provide it with future messages.

**Where**
 Extension method for hooking in a `WhereObserver` which allows for filtering on `Msg`.

**Select**
Extension method for hooking in a `MapObserver` which allows for mapping/transforming from `Msg` to `T`.

**Sample**
```csharp
using System;
using System.Linq;
using System.Threading;
using System.Threading.Tasks;
using NATS.Client;
using NATS.Client.Rx;
using NATS.Client.Rx.Ops; //Can be replaced with using System.Reactive.Linq;

namespace RxSample
{
    class Program
    {
        static void Main(string[] args)
        {
            using (var cn = new ConnectionFactory().CreateConnection())
            {
                var temperatures =
                    cn.Observe("temperatures")
                        .Where(m => m.Data?.Any() == true)
                        .Select(m => BitConverter.ToInt32(m.Data, 0));

                temperatures.Subscribe(t => Console.WriteLine($"{t}C"));

                temperatures.Subscribe(t => Console.WriteLine($"{(t * 9 / 5) + 32}F"));

                var cts = new CancellationTokenSource();

                Task.Run(async () =>
                {
                    var rnd = new Random();

                    while (!cts.IsCancellationRequested)
                    {
                        cn.Publish("temperatures", BitConverter.GetBytes(rnd.Next(-10, 40)));

                        await Task.Delay(1000, cts.Token);
                    }
[...]

Issue #149, opened 2017-02-28
Rx support


jaklithn:
Daniel Wertheim has implemented a very nice Rx support in his .Net client. https://github.com/danielwertheim/mynatsclient
I really like that approach as it is fast, elegant and modern. On the other hand I prefer to use officially supported libraries as they tend over time to be more stable and supported.
Can you please consider to include the Rx support in the core of the official client lib?

ColinSullivan1:
@jaklithn , Thank you for looking at NATS, and raising this issue. Absolutely, and I certainly agree - Rx is a nice programming model for the reasons you mentioned.  Providing Rx support is on our list; and I hope to get to it fairly soon.  

In the meantime, we also do accept user contributions -  a performant, well written, separate Rx assembly over the official NATS assembly is something we would certainly consider adopting.

nataly87s:
I want to work on this issue

ColinSullivan1:
@nataly87s , Absolutely, would love the help here!  

I'd suggest as next steps:

1)  You propose an API, which we can review.  A simple layer on top of this NATS assembly (versus built-into NATS) would be ideal.  It'd be be easiest if you'd provide a skeleton API with documentation in a repository you own, where we can collaborate.

2) After the API is defined, once code is complete and unit tests are passing, I'll setup a NATS repository, setup CI, get the NuGet package put together, etc.  You can create a PR with all of your code so github properly reflects your contribution.

3) We'll get some members of NATS community to review it and test it.

4) We create an alpha release, and move forward with your contribution.

Throughout this process I'll be available to answer questions you may have, review code, etc.

We're pretty flexible and want to make things easy, so if you prefer to work another way please let me know.

BTW - If you would like to see what a third party NATS Rx API looks like, check out https://github.com/danielwertheim/mynatsclient (as mentioned above).

Sound good?  Looking forward to your contribution - Thanks!

nataly87s:
I think the easiest way to do it would be to just add a project to this repository with extention methods for `IConnection`. 
Of course I'll add documentation and tests as well

ColinSullivan1:
If it facilitates development I'm OK starting that way, but we'll want to differentiate releases for NATS and NATS-RX, which will require different repositories.

nataly87s:
It was easier for me to just rewrite everything, so I created a new client
https://github.com/nataly87s/SimpleNatsClient
I still need to add reconnect, and add some more integration tests, but what do you think so far?

[...]

PR #304, opened 2019-10-15
Adds Observable support

danielwertheim:
This PR adds observable support as defined in #303

### DIFF
--- a/src/NATS.Client/Rx/INATSObservable.cs
+++ b/src/NATS.Client/Rx/INATSObservable.cs
@@ -1,0 +1,24 @@
+﻿// Copyright 2015-2018 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace NATS.Client.Rx
+{
+    /// <summary>
+    /// Marker interface to have shipped extension methods to not interfear
+    /// with offical RX extensions.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public interface INATSObservable<out T> : IObservable<T>, IDisposable { }
+}
--- a/src/NATS.Client/Rx/INATSObservable.cs
+++ b/src/NATS.Client/Rx/INATSObservable.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2015-2018 The NATS Authors
+﻿// Copyright 2019 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
--- a/src/NATS.Client/Rx/INATSObservable.cs
+++ b/src/NATS.Client/Rx/INATSObservable.cs
@@ -16,8 +16,8 @@ using System;
 namespace NATS.Client.Rx
 {
     /// <summary>
-    /// Marker interface to have shipped extension methods to not interfear
-    /// with offical RX extensions.
+    /// Marker interface to have shipped extension methods to not interfere
+    /// with official RX extensions.
     /// </summary>
     /// <typeparam name="T"></typeparam>
     public interface INATSObservable<out T> : IObservable<T>, IDisposable { }
--- a/src/NATS.Client/Rx/NATSObservable.cs
+++ b/src/NATS.Client/Rx/NATSObservable.cs
@@ -1,0 +1,129 @@
+﻿// Copyright 2015-2018 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NATS.Client.Rx
+{
+    public abstract class NATSObservable<T> : INATSObservable<T>
+    {
+        private readonly ConcurrentDictionary<int, ObserverSubscription> subscriptions = new ConcurrentDictionary<int, ObserverSubscription>();
+
+        protected void InvokeObservers(T data)
+        {
+            foreach (var sub in subscriptions.Values)
+            {
+                try
+                {
+                    sub.Observer.OnNext(data);
+                }
+                catch
+                {
+                    //Missbehaving observer will remove its subscription
+                    //and dispose it. No invoke of OnError as it's not the
+                    //producer that is failing.
+                    subscriptions.TryRemove(sub.Id, out _);
+
+                    try
+                    {
+                        sub.Dispose();
+                    }
+                    catch
+                    {
+                        // ignored
+                    }
+                }
+            }
+        }
+
+        private void DisposeSubscription(ObserverSubscription sub)
+        {
+            if(!subscriptions.TryRemove(sub.Id, out _))
+                return;
+
+            try
+            {
+                sub.Observer.OnCompleted();
+            }
+            catch
+            {
+                // ignored
+            }
+        }
+
+        public IDisposable Subscribe(IObserver<T> observer)
+        {
+            if (observer == null)
+                throw new ArgumentNullException(nameof(observer));
+
+            var sub = new ObserverSubscription(observer, DisposeSubscription);
+            
+            if(!subscriptions.TryAdd(sub.Id, sub))
+                throw new ArgumentException("Could not subscribe observer. Ensure it has not already been subscribed.", nameof(observer));
+
+            return sub;
+        }
+
+        private sealed class ObserverSubscription : IDisposable
+        {
+            private readonly Action<ObserverSubscription> disposer;
+
+            public readonly int Id;
+            public readonly IObserver<T> Observer;
+
+            public ObserverSubscription(IObserver<T> observer, Action<ObserverSubscription> disposer)
+            {
+                Id = observer.GetHashCode();
+                Observer = observer;
+                this.disposer = disposer;
+            }
+
+            public void Dispose() => disposer(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if(!disposing)
+                return;
+
+            var exceptions = new List<Exception>();
+
+            foreach (var obSub in subscriptions)
+            {
+                try
+                {
+                    subscriptions.TryRemove(obSub.Key, out _);
+                    obSub.Value.Dispose();
+                }
+                catch (Exception ex)
+                {
+                    exceptions.Add(ex);
+                }
+            }
+
+            if(exceptions.Any())
+                throw new NATSRxException(
+                    "One or more exception(s) occurred while disposing the NATS Observable's' subscriptions. See inner exception (AggregateException) for more details.",
+                    new AggregateException(exceptions));
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+    }
+}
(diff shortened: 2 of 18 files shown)


src/NATS.Client/Rx/INATSObservable.cs
```diff
@@ -0,0 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace NATS.Client.Rx
+{
+    /// <summary>
+    /// Marker interface to have shipped extension methods to not interfear
```
watfordgnf:
```suggestion
    /// Marker interface to have shipped extension methods to not interfere
```

src/NATS.Client/Rx/INATSObservable.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// with offical RX extensions.
```
watfordgnf:
```suggestion
    /// with official RX extensions.
```

src/NATS.Client/Rx/NATSObservable.cs
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        public IDisposable Subscribe(IObserver<T> observer)
+        {
+            if (observer == null)
+                throw new ArgumentNullException(nameof(observer));
+
+            var sub = new ObserverSubscription(observer, DisposeSubscription);
+            
+            if(!subscriptions.TryAdd(sub.Id, sub))
```
watfordgnf, line 74: Should we consider returning the existing subscription if this is called twice?

src/NATS.Client/Rx/Ops/DelegatingObserver.cs
```diff
@@ -0,0 +1,43 @@
...
+        private readonly Action<T> onNext;
+        private readonly Action<Exception> onError;
+        private readonly Action onCompleted;
+
+        internal DelegatingObserver(
+            Action<T> onNext,
+            Action<Exception> onError = null,
+            Action onCompleted = null)
+        {
+            this.onNext = onNext;
```
watfordgnf:
```suggestion
            this.onNext = onNext ?? throw new ArgumentNullException(nameof(onNext));
```

src/NATS.Client/Rx/Ops/SafeObserver.cs
```diff
@@ -0,0 +1,52 @@
...
+        private readonly Action<T> onNext;
+        private readonly Action<Exception> onError;
+        private readonly Action onCompleted;
+
+        internal SafeObserver(
+            Action<T> onNext,
+            Action<Exception> onError = null,
+            Action onCompleted = null)
+        {
+            this.onNext = onNext;
```
watfordgnf:
```suggestion
            this.onNext = onNext ?? throw new ArgumentNullException(nameof(onNext));
```

src/NATS.Client/Rx/NATSObservable.cs
```diff
@@ -0,0 +1,129 @@
...
+        private sealed class ObserverSubscription : IDisposable
+        {
+            private readonly Action<ObserverSubscription> disposer;
+
+            public readonly int Id;
+            public readonly IObserver<T> Observer;
+
+            public ObserverSubscription(IObserver<T> observer, Action<ObserverSubscription> disposer)
+            {
+                Id = observer.GetHashCode();
```
watfordgnf, line 89: We're using hash codes for uniqueness, however, someone could conceptually make their own and our invariant would be broken. We're looking for unique-by-reference-equality, right?

src/NATS.Client/Rx/Ops/MapObservable.cs
```diff
@@ -0,0 +1,54 @@
...
+        public MapObservable(INATSObservable<TSrc> src, Func<TSrc, TResult> predicate)
+        {
+            this.src = src ?? throw new ArgumentNullException(nameof(src));
+            this.mapper = predicate ?? throw new ArgumentNullException(nameof(predicate));
+        }
+
+        public void Dispose() => src?.Dispose();
+
+        public IDisposable Subscribe(IObserver<TResult> observer)
+            => src.SubscribeSafe(new MapObserver(observer, mapper));
```
watfordgnf, line 32: Does this need a test for `observer` against `null`?

src/NATS.Client/Rx/Ops/WhereObservable.cs
```diff
@@ -0,0 +1,58 @@
...
+        public WhereObservable(INATSObservable<T> src, Func<T, bool> predicate)
+        {
+            this.src = src ?? throw new ArgumentNullException(nameof(src));
+            this.predicate = predicate ?? throw new ArgumentNullException(nameof(predicate));
+        }
+
+        public void Dispose() => src?.Dispose();
+
+        public IDisposable Subscribe(IObserver<T> observer)
+            => src.SubscribeSafe(new WhereObserver(observer, predicate));
```
watfordgnf, line 32: Does this need a test for `observer` against `null`?

src/NATS.Client/Rx/RxExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <summary>
+        /// Subscribes to the passed subject and returns a hot observable. Hence unless you
+        /// subscribe to the observable, no message will be handled and old messages will
+        /// not be delivered, only new messages.
+        /// </summary>
+        /// <param name="cn"></param>
+        /// <param name="subject"></param>
+        /// <returns></returns>
+        public static INATSObservable<Msg> Observe(this IConnection cn, string subject)
+            => cn.SubscribeAsync(subject).ToObservable();
```
watfordgnf: We should probably test `cn` against `null`.

watfordgnf (commented):
Some general comments, liking the way this looks!

src/NATS.Client/Rx/NATSObservable.cs
danielwertheim: Just as if we would add the observer to e.g. a HashSet. We could switch for a GUID if you feel it is that important?

src/NATS.Client/Rx/NATSObservable.cs
danielwertheim: But why should you call it twice?

src/NATS.Client/Rx/Ops/DelegatingObserver.cs
danielwertheim: Completely missed I could accept suggestion. It's been fixed though

src/NATS.Client/Rx/Ops/SafeObserver.cs
danielwertheim: Fixed

src/NATS.Client/Rx/NATSObservable.cs
watfordgnf: Perhaps this is a non-issue if users would never logically pass the same observer twice to Subscribe.

src/NATS.Client/Rx/Ops/MapObservable.cs
danielwertheim: The `MapObserver` now requires it.

src/NATS.Client/Rx/Ops/WhereObservable.cs
danielwertheim: The WhereObserver now requires it.

src/NATS.Client/Rx/RxExtensions.cs
danielwertheim: We now check so that it isn't null.

src/NATS.Client/Rx/NATSObservable.cs
danielwertheim: The idea though, is that you should not add the exact same observer twice, hence e.g. a GUID would not work as I see it.

danielwertheim:
This would also solve #149

src/NATS.Client/Rx/NATSObservable.cs
watfordgnf: What would be the downside to making the key simply the observer itself? We'd retain HashSet semantics and avoid the odd situation where two equal hash codes could occur for two distinct objects.

danielwertheim: I guess it's just habit of as much as possible using a value object as a key due to allocations etc.

Using the "object" as the key will call it's GetHashCode when adding and GetHashCode + Equals e.g when getting so I'm not sure how that would solve the issue you want to solve.

src/NATS.Client/Rx/INATSObservable.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright 2015-2018 The NATS Authors
```
ColinSullivan1: Minor thing...  Could you update the license headers to be `Copyright 2019 The NATS Authors`?

src/NATS.Client/Rx/NATSObservable.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            foreach (var sub in subscriptions.Values)
+            {
+                try
+                {
+                    sub.Observer.OnNext(data);
+                }
+                catch
+                {
+                    //Missbehaving observer will remove its subscription
```
ColinSullivan1: Typo `Misbehaving`.

src/NATS.Client/Rx/NATSObservableSubscription.cs
```diff
@@ -0,0 +1,47 @@
...
+    public sealed class NATSObservableSubscription : NATSObservable<Msg>, IDisposable
+    {
+        private readonly IAsyncSubscription subscription;
+
+        private NATSObservableSubscription(IAsyncSubscription subscription)
+        {
+            this.subscription = subscription ?? throw new ArgumentNullException(nameof(subscription));
+
+            this.subscription.MessageHandler += OnIncomingMessage;
+            this.subscription.Start();
```
ColinSullivan1, line 27: I'm wondering if it'd be useful (or feasible) to also setup an AsyncErrorHandler (normally set in [options](https://github.com/nats-io/nats.net/blob/master/src/NATS.Client/Options.cs#L72)) to intercept async errors from the server and invoke `onError` in the observable.  I know these errors are attached to a connection today, but it might be useful to surface slow consumer errors through `onError`.  wdyt?

src/Tests/IntegrationTests/TestRx.cs
```diff
@@ -0,0 +1,242 @@
...
+                        sync.WaitForAll();
+                        Assert.Equal(payload1, interceptingOb1.OnNextResults.Single());
+                        Assert.Equal(payload2, interceptingOb2.OnNextResults.Single());
+                    }
+                }
+            }
+        }
+
+        [Fact]
+        public void WhenDispoingAnObserverItShouldNotReceiveMoreMessages()
```
ColinSullivan1: typo...

src/Tests/IntegrationTests/TestRx.cs
```diff
@@ -0,0 +1,242 @@
...
+
+                        Assert.Equal(new[] { payload1.Data, payload2.Data }, interceptingOb1.OnNextResults.Select(i => i.Data).ToArray());
+                        Assert.Equal(new[] { payload1.Data }, interceptingOb2.OnNextResults.Select(i => i.Data).ToArray());
+                    }
+                }
+            }
+        }
+
+        [Fact]
+        public void WhenDispoingAnObservableNoMoreDispatchesShouldBeDone()
```
ColinSullivan1: Typo...

ColinSullivan1 (changes requested):
This looks great!  Some minor comments/typos and a question regarding slow consumers.  Could you add some brief example code into the README?

It doesn't need to be part of this PR (can be separate), but an example project users could reference would go a long way to helping them started with this API.

src/NATS.Client/Rx/NATSObservableSubscription.cs
danielwertheim: Calling `OnError` would terminate the observer, meaning if the connection comes up again, the subscribed observer would not get any further dispatches. I don't think that is optimal.

src/NATS.Client/Rx/INATSObservable.cs
danielwertheim: Updated the ones added in this PR.

src/Tests/IntegrationTests/TestRx.cs
danielwertheim: Fixed

src/Tests/IntegrationTests/TestRx.cs
danielwertheim: Fixed

src/NATS.Client/Rx/NATSObservable.cs
danielwertheim: Fixed

danielwertheim:
@ColinSullivan1 Added `RxSample` project under Samples and added code to README,

src/NATS.Client/Rx/NATSObservableSubscription.cs
ColinSullivan1: Ahh, good point.  I agree - thanks.

ColinSullivan1 (approved):
LGTM.  Thanks - well done!

ColinSullivan1:
I think this is ready to be merged.  @watfordgnf , wdyt?

watfordgnf (approved):

PR #305, opened 2019-10-17
Fixes clean up of linked cancellation token source

danielwertheim:
The `InflightRequest` construct now also takes care of the linked cancellation token source. I've also updated so that `Dispose` actually is called where used.


watfordgnf (approved):

ColinSullivan1 (approved):
LGTM.

PR #306, opened 2019-10-17
Uses lock was taken logic in channel

danielwertheim:
The construct as `lock` would use (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/lock-statement)

Could potentially look at replacing the code with a `BlockingCollection` wrapping a `ConcurrentQueue` but that would be another PR.


watfordgnf (approved):

ColinSullivan1:
In the past attempting to use the concurrent structures really affected performance likely due to internal additional locking we didn't need.  But things could have changed....

ColinSullivan1 (approved):
LGTM.

PR #308, opened 2019-10-29
Test with .NET core 3.0

ColinSullivan1:

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -97,6 +97,14 @@ stages:
         projects: 'src/Tests/**/IntegrationTests.csproj'
         arguments: '-c $(BuildConfiguration) -f netcoreapp2.2 --no-build'
         testRunTitle: 'IntegrationTests .NetCoreApp2.2'
+
+    - task: DotNetCoreCLI@2
+      displayName: 'IntegrationTests .NetCoreApp3.0'
+      inputs:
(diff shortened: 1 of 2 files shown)


danielwertheim (commented):
The tests will not run under 3.0 in the pipeline unless the `azure-pipelines.yml` file is extended. It specifically runs 4.5.2 and netcoreapp2.2.

But do we need both netcoreapp2.2 and 3.0 or is switching to 3.0 enough?

I pushed the required changes.

watfordgnf:
We've heard from a few folks that there are certain issues on Linux that differ between 2.1, 2.2, and 3.0, so it may be beneficial to run more than one leg on more than one OS.

watfordgnf (approved):

ColinSullivan1:
Thanks for the reviews!  I'd prefer to run netcore2.2 and netcore3.0, just to make sure we're covered - thanks for pushing the changes @danielwertheim.  

@watfordgnf, I agree, and good idea - we're seeing a lot of discrepancies and bugs in .NET core on linux in code that has been running bug-free for years, particularly around socket handling (connecting, keep-alives, etc).  I'd prefer to handle that in a separate PR ad first understand the scope of the required changes for a clean build.

I'm getting concerned about running on linux in mission critical systems, and am wondering if we need to make a statement.

PR #311, opened 2019-10-29
Removes duplicate call to kickFlusher

danielwertheim:
`setFlusherDone(true)` already causes `kickFlusher` to run. Or is this done to be "sure" `kickFlusher` is called e.g. if someone refactors something that isn't caught by tests?


watfordgnf (approved):

ColinSullivan1:
I'm fairly sure this was a mistake, but let me look at the blame logs. ;)

EDIT:  Looks like we lost history in moving files around, but no worries there.

ColinSullivan1 (approved):
LGTM.

Issue #268, opened 2019-06-10
Reconnect can throw an unhandled exception


harrisa1:
During regular usage there was something that caused the nats connection to drop (i.e. a network blip).  After this the client tried to reconnect and when it did it hit an exception
```
Type=System.InvalidOperationException
Source=System.Collections
Message=Collection was modified; enumeration operation may not execute.
StackTrace:
   at System.Collections.Generic.Queue`1.Enumerator.MoveNext()
   at NATS.Client.Connection.clearPendingFlushCalls()
   at NATS.Client.Connection.doReconnect()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
```
Looking at the above I am guessing that a ConcurrentQueue would probably serve better since the collection can be modified while enumerating.

watfordgnf:
`pongs` appears to be covered by the `mu` lock. It would be good to know how `pongs` was altered outside of `mu`.

Found it:
https://github.com/nats-io/nats.net/blob/012ce23603b384dd422735e33b10f12c8e3309e7/NATS.Client/Conn.cs#L3557-L3562

`removeFlushEntry` should be protected by `mu` but is not.

RobinsonWM:
We encountered a crash today that I believe has the same root cause but a different exception.  I wanted to paste it here in case anyone else is searching for this issue.
It is in the same method (`clearPendingFlushCalls()`) as the OP's issue, and it is also caused by access to the `pongs` queue while another thread is mutating it, just from a different line.

```no-highlight
Type=System.IndexOutOfRangeException
Source=System.Private.CoreLib
Message=Index was outside the bounds of the array.
StackTrace:
   at System.Array.Clear(Array array, Int32 index, Int32 length)
   at System.Collections.Generic.Queue`1.Clear()
   at NATS.Client.Connection.clearPendingFlushCalls()
   at NATS.Client.Connection.doReconnect()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
```

ColinSullivan1:
@RobinsonWM , thank you for updating the issue.  This will be addressed soon...

[...]

PR #312, opened 2019-10-29
Uses ConcurrentQueue instead of manual locking for pongs

danielwertheim:
Uses `ConcurrentQueue` and `Interlocked` instead of manual locking which e.g. should solve #268

Regarding performance: https://docs.microsoft.com/en-us/dotnet/standard/collections/thread-safe/when-to-use-a-thread-safe-collection#concurrentqueuet-vs-queuet

Did some testing with Benchmark DotNet having parallel producer and consumer working with a `lock` and normal `Queue` and one with no locks and `ConcurrentQueue` and didn't see any real difference in speed as well as memory consumption: https://gist.github.com/danielwertheim/23ca65dc73cf6c4f24e6241c340b49b0

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -3494,9 +3494,6 @@ namespace NATS.Client
         // call outstanding and we call close.
         private bool removeFlushEntry(SingleUseChannel<bool> chan)
         {
-            if (pongs.Count == 0)
-                return false;
-
             if (!pongs.TryDequeue(out var start))
                 return false;
 
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -142,7 +142,7 @@ namespace NATS.Client
         private ConcurrentDictionary<Int64, Subscription> subs = 
             new ConcurrentDictionary<Int64, Subscription>();
         
-        private Queue<SingleUseChannel<bool>> pongs = new Queue<SingleUseChannel<bool>>();
+        private readonly ConcurrentQueue<SingleUseChannel<bool>> pongs = new ConcurrentQueue<SingleUseChannel<bool>>();
 
         internal MsgArg   msgArgs = new MsgArg();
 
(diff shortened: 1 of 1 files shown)


watfordgnf:
If I remember the discussions during netcore's development, the particular performance pattern of ConcurrentQueue under .NET Framework was geared towards multithreaded performance, and did not do as well with single producer - single consumer. It is quite likely this has shifted.

src/NATS.Client/Conn.cs
```diff
@@ -3511,14 +3494,13 @@ internal virtual void removeSub(Subscription s)
         // call outstanding and we call close.
         private bool removeFlushEntry(SingleUseChannel<bool> chan)
         {
-            if (pongs == null)
+            if (pongs.Count == 0)
                 return false;
 
-            if (pongs.Count == 0)
+            if (!pongs.TryDequeue(out var start))
                 return false;
```
watfordgnf: I think it would be easier to reason if there was only one call, rather than two calls:
```csharp
            if (!pongs.TryDequeue(out var start))
                return false;
```
(that being said, I don't believe having both introduces any races)

src/NATS.Client/Conn.cs
```diff
@@ -916,9 +915,7 @@ private void pingTimerCallback(object state)
                     return;
                 }
 
-                pout++;
-
-                if (pout > Opts.MaxPingsOut)
+                if (Interlocked.Increment(ref pout) > Opts.MaxPingsOut)
```
watfordgnf, line 918: `Interlocked.Increment` returns the original value, so this should change to `>= Opts.MaxPingsOut` to retain the original behavior.

watfordgnf (approved):
Two comments, otherwise looks good.

src/NATS.Client/Conn.cs
danielwertheim: > **Returns** "The incremented value."
https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.increment?view=netframework-4.8

src/NATS.Client/Conn.cs
danielwertheim: Fixed

src/NATS.Client/Conn.cs
danielwertheim: @watfordgnf Given the docs, do you still want a change?

watfordgnf: No, that's my bad. Confused that with Interlocked.CompareExchange.

ColinSullivan1 (approved):
LGTM.

PR #284, opened 2019-08-30
Add Reconnect Buffering

ColinSullivan1:
Implement reconnect buffer sizing and disabling, a feature available in the golang client.

Resolves #251.

Signed-off-by: Colin Sullivan <colin@synadia.com>

PR #314, opened 2019-10-30
Manually redo of previous reconnect-buffer PR

### LINKED ISSUE #251: disable sending message during reconnect
hello

how can i disable sending messages after reconnect  that are created during connection lost 

currently published messages are buffered when there is no connection with server and  send immediately after the successful re connection

like this one
https://github.com/nats-io/java-nats/issues/138

danielwertheim:
It was easier to manually reapply changes from PR #284 Also fixes comments in that review. I've opened this PR up as a new PR so that you can compare that I didn't miss anything.

As it replaces #284 it also resolves #251

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -2435,7 +2435,7 @@ namespace NATS.Client
                         else
                             kickFlusher();
 
-                        if (pending.Position + count + pubProtoLen > rbsize)
+                        if (bw.Position + count + pubProtoLen > rbsize)
                             throw new NATSReconnectBufferException("Reconnect buffer exceeded.");
                     }
                 }
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -2419,6 +2419,28 @@ namespace NATS.Client
                 // write our pubProtoBuf buffer to the buffered writer.
                 int pubProtoLen = writePublishProto(pubProtoBuf, subject, reply, count);
 
+                // Check if we are reconnecting, and if so check if
+                // we have exceeded our reconnect outbound buffer limits.
+                // Don't use IsReconnecting to avoid locking.
+                if (status == ConnState.RECONNECTING)
+                {
+                    int rbsize = opts.ReconnectBufferSize;
+                    if (rbsize != 0)
+                    {
+                        if (rbsize == -1)
+                        {
+                            throw new NATSReconnectBufferException("Reconnect buffering has been disabled.");
+                        }
+
+                        bw.Flush();
+
+                        if (pending.Position + count + pubProtoLen > rbsize)
+                        {
+                            throw new NATSReconnectBufferException("Reconnect buffer exceed.");
+                        }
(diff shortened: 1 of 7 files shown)


src/NATS.Client/Conn.cs
```diff
@@ -2419,6 +2419,28 @@ internal void publish(string subject, string reply, byte[] data, int offset, int
...
+                {
+                    int rbsize = opts.ReconnectBufferSize;
+                    if (rbsize != 0)
+                    {
+                        if (rbsize == -1)
+                        {
+                            throw new NATSReconnectBufferException("Reconnect buffering has been disabled.");
+                        }
+
+                        bw.Flush();
```
danielwertheim: @ColinSullivan1 should this take the flag into consideration?

ColinSullivan1: Yes, it should, good catch.

src/NATS.Client/Conn.cs
```diff
@@ -2419,6 +2419,28 @@ internal void publish(string subject, string reply, byte[] data, int offset, int
...
+
+                        if (pending.Position + count + pubProtoLen > rbsize)
+                        {
+                            throw new NATSReconnectBufferException("Reconnect buffer exceed.");
```
ColinSullivan1: exceed -> exceeded

danielwertheim: Fixed

src/NATS.Client/Conn.cs
danielwertheim: Fixed

ColinSullivan1 (approved):
LGTM - Thanks!

danielwertheim:
@ColinSullivan1 fixed the failing test but I wonder if the "fix" is OK as in, is it OK for it to fail due to increased payload or was the intention to ensure the "queue" is filling up?d

ColinSullivan1:
Hmm...  I wanted to make sure the 2nd publish failed as it would overrun the full buffer, but `PUB foo 20\r\n<20 bytes>\r\n` is 34 bytes, which should have thrown on the first publish?

src/NATS.Client/Conn.cs
```diff
@@ -2419,6 +2419,27 @@ internal void publish(string subject, string reply, byte[] data, int offset, int
...
+                    {
+                        if (rbsize == -1)
+                            throw new NATSReconnectBufferException("Reconnect buffering has been disabled.");
+
+                        if (flushBuffer)
+                            bw.Flush();
+                        else
+                            kickFlusher();
+
+                        if (pending.Position + count + pubProtoLen > rbsize)
```
danielwertheim: @ColinSullivan1  Replying to your Q here instead to get the context.

First publish: `0+20+12 > 32`

Second publish: `0+21+12 > 32`

ColinSullivan1: LGTM.  Thanks!

ColinSullivan1:
LGTM.  Thanks!

PR #316, opened 2019-11-04
Resolves #272 by parsing out IpV4 or IpV6

### LINKED ISSUE #272: Nats.Client can't connect to IPv6 server on .net core < 3.0
Hello. I'm facing problem with connecting to nats server via ipv6. 

```c#
var cf = new ConnectionFactory();
var c = cf.CreateConnection("nats://[::1]:4222");
```
it crashes with exception
```
Unhandled Exception: NATS.Client.NATSNoServersException: Unable to connect to a server.
   at NATS.Client.Connection.connect() in C:\Users\Sonic\RiderProjects\NatsIPv6\NatsIPv6\NATS.Client\Conn.cs:line 1131
   at NATS.Client.ConnectionFactory.CreateConnection(Options opts) in C:\Users\Sonic\RiderProjects\NatsIPv6\NatsIPv6\NATS.Client\ConnectionFactory.cs:line 171
   at NATS.Client.ConnectionFactory.CreateConnection(String url) in C:\Users\Sonic\RiderProjects\NatsIPv6\NatsIPv6\NATS.Client\ConnectionFactory.cs:line 48
   at NatsIPv6.Program.Main(String[] args) in C:\Users\Sonic\RiderProjects\NatsIPv6\NatsIPv6\Program.cs:line 14
```
As i understand, problem is here - https://github.com/nats-io/nats.net/blob/master/NATS.Client/Conn.cs#L427. By default TCPClient created with AddressFamily.InterNetwork which is IPv4 only. If i change this line to 
```c#
client = new TcpClient(AddressFamily.InterNetworkV6);
```
than it can connect to ipv6 nats without problem.

More interestingly, everything works fine without any modifications on .net core 3.0, cause of this pr - https://github.com/dotnet/corefx/pull/30036. 

But i think it should be fixed for 2.2 also. Best way to do it and stay compatible with 3.0 changes is to resolve ip address before tcpclient creation. Something like that:
```c#
var ipAddress = Dns.GetHostAddresses(s.url.Host).FirstOrDefault();
if (ipAddress == null)
{
     throw new NATSConnectionException("can't resolve server ip");
}
client = new TcpClient(ipAddress.AddressFamily);
var task = client.ConnectAsync(ipAddress, s.url.Port);
``` 

This way it works for ipv4 and ipv6 on both .net core 2.2 and 3.0. If this is correct approach - i can make pr.

danielwertheim:
Explicitly looking for IpV4 and IpV6 otherwise falling back to framework support to not change existing behavior. The goal was to change as little as possible to solve the issue. We could of course to the DNS resolve or add connection option if we want. But that could int that case perhaps be added as a separate PR as that would be more of a "feature".

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -24,6 +24,7 @@ using System.Security.Cryptography.X509Certificates;
 using System.Security.Authentication;
 using System.Globalization;
 using System.Diagnostics;
+using NATS.Client.Extensions;
 using NATS.Client.Internals;
 
 namespace NATS.Client
@@ -439,7 +440,7 @@ namespace NATS.Client
                         sslStream = null;
                     }
 
-                    client = new TcpClient();
+                    client = createTcpClient(s.url);
                     var task = client.ConnectAsync(s.url.Host, s.url.Port);
                     // avoid raising TaskScheduler.UnobservedTaskException if the timeout occurs first
                     task.ContinueWith(t => GC.KeepAlive(t.Exception), TaskContinuationOptions.OnlyOnFaulted);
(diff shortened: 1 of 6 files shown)


watfordgnf (approved):

ColinSullivan1 (approved):
LGTM - thanks!

Issue #222, opened 2018-07-11
Needs to set timeout to StreamReader


ys-zhao:
When Nats server is stuck, sr.ReadLine() won't return, needs to pass Options.Timeout to readOp()

https://github.com/nats-io/csharp-nats/blob/d1b1ec379a865ebafc742e0707dc57871fbb7ff6/NATS.Client/Conn.cs#L1298

ColinSullivan1:
@ys-zhao , Thank you for raising this issue, I'll take a look.

danielwertheim:
@ys-zhao @ColinSullivan1 How would one re-produce this in a test? I mean, it's rather "simple" to add an `option.ReadTimeout` or similar that could be configured, but I'm wondering when this happens.

ColinSullivan1:
The only deterministic way would be to mock the server, which imo is not worth the time.  I think some simple tests setting it and testing bad values will suffice.

PR #318, opened 2019-11-05
Adds option.ReadTimeout

danielwertheim:
Solves #222 by adding support for configuring `option.ReadTimeout`.

### DIFF
--- a/src/Tests/IntegrationTests/TestConnection.cs
+++ b/src/Tests/IntegrationTests/TestConnection.cs
@@ -458,7 +458,7 @@ namespace IntegrationTests
         {
             var opts = Context.GetTestOptionsWithDefaultTimeout(Context.Server1.Port);
             opts.AllowReconnect = false;
-            opts.HandshakeReadTimeout = 500;
+            opts.Timeout = 500;
 
             using (var s = NATSServer.CreateFastAndVerify(Context.Server1.Port))
             {
--- a/src/Tests/IntegrationTests/TestConnection.cs
+++ b/src/Tests/IntegrationTests/TestConnection.cs
@@ -453,6 +453,93 @@ namespace IntegrationTests
             }
         }
 
+        [Fact]
+        public void WhenReadTimeoutOccursAndNoReconnectItShouldGetDisconnected()
+        {
+            var msgEv = new AutoResetEvent(false);
+            var disconnectedEv = new AutoResetEvent(false);
+            var reconnectedEv = new AutoResetEvent(false);
+            
+            var opts = Context.GetTestOptionsWithDefaultTimeout(Context.Server1.Port);
+            opts.AllowReconnect = false;
+            opts.ReadTimeout = 500;
+
+            opts.DisconnectedEventHandler = (sender, args) => disconnectedEv.Set();
+            opts.ReconnectedEventHandler = (sender, args) => reconnectedEv.Set();
+
+            const string subject = "1f9ba82b90474c37a519d97ea12a7f22";
+            var payload = new byte[0];
+
+            using (var s = NATSServer.CreateFastAndVerify(Context.Server1.Port))
+            {
+                using (var cn = Context.ConnectionFactory.CreateConnection(opts))
+                {
+                    cn.SubscribeAsync(subject, (sender, args) =>
+                    {
+                        msgEv.Set();
+                        Thread.Sleep(opts.ReadTimeout.Value + 250); //Causes time to pass so that ReadTimeout occurs
+                    });
+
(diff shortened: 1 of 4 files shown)


src/Tests/IntegrationTests/TestConnection.cs
```diff
@@ -453,6 +453,93 @@ public void TestInfineReconnect()
...
+            var payload = new byte[0];
+
+            using (var s = NATSServer.CreateFastAndVerify(Context.Server1.Port))
+            {
+                using (var cn = Context.ConnectionFactory.CreateConnection(opts))
+                {
+                    cn.SubscribeAsync(subject, (sender, args) =>
+                    {
+                        msgEv.Set();
+                        Thread.Sleep(opts.ReadTimeout.Value + 250); //Causes time to pass so that ReadTimeout occurs
```
watfordgnf: If I'm understanding this test correctly, with the Read Timeout set, any busywork on the message handler can cause this to occur?

If that is the case, should we provide guidance in the `ReadTimeout` documentation?

danielwertheim: Any suggestion on what you would like the docs to be?

What we do is to affect the `TcpClient.ReceiveTimeout` which in turns sets the underlying `stream.ReadTimeout`.

https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.tcpclient.receivetimeout?view=netframework-4.8#System_Net_Sockets_TcpClient_ReceiveTimeout

I was a bit surprised that I could make "time pass" in that sense in the test, but it's due to the fact that there's nothing inbound during that time as no new publish is made and no ping.

danielwertheim:
We really need to take a look at some flaky tests... Successful build manually provided here (as Azure DevOps has a known issue with re-triggering builds) https://dev.azure.com/NATS-CI/NATS-CI/_build/results?buildId=99&view=results

src/Tests/IntegrationTests/TestConnection.cs
watfordgnf: I'm surprised as well, I guess we'd want remarks letting folks know they should keep all processing to a minimum in their async handlers.

Does this affect using a synchronous message loop?

danielwertheim: I see this option as something "advanced users" might want to tweak. And in general all handlers should be "short", regardless of this PR or not. Otherwise they still risk getting cut of as a slow consumer, right? Given this, what is there more to change in this PR to complete it?

danielwertheim:
Will that not be a bit confusing as we already have a timeout for `Connect` https://github.com/nats-io/nats.net/blob/master/src/NATS.Client/Conn.cs#L447

I followed the "pattern" that is in place for `processExpectedInfo` https://github.com/nats-io/nats.net/blob/master/src/NATS.Client/Conn.cs#L1194

Setting and reseting the `conn.ReceiveTimeout` during the handshake.

Could do something similar as we do with the send timeout. E.g.

```csharp
if (opts.HandshakeTimeout.HasValue)
{
    var completedInTime = Task.Run(() =>
    {
        processExpectedInfo();
        sendConnect();
    }).Wait(opts.HandshakeTimeout.Value);
    if(!completedInTime)
        throw new NATSConnectionException("Timeout during server handshake.");
}
else
{
    processExpectedInfo();
    sendConnect();
}
```

ColinSullivan1:
@danielwertheim , good point, that would be confusing.... that pattern looks good.

ColinSullivan1:
With the proposed changes, I think the current `Options.Timeout` property would suffice, and we could avoid adding two new properties.  It wouldn't be 100% accurate if this edge case is hit, but close enough imo.  wdyt?

danielwertheim:
But the Timeout is used on the outer scope, right? Thought the idea was to be able to tweak the read side in the connect?

ColinSullivan1:
For most projects, yeah, that'd be a nice to have, but with NATS I'd really like to keep the API footprint as small as possible.  The original issue creator just wanted a timeout, not necessarily to specifically set it... imo just reusing the Timeout is reasonable considering this is a fairly extreme corner case.  I'd be worried users without a lot of networking background could get confused if there are too many timeouts.

danielwertheim:
@ColinSullivan1 Fixed.

src/Tests/IntegrationTests/TestConnection.cs
danielwertheim: Not relevant anymore

ColinSullivan1 (approved):
LGTM - Thanks!

PR #321, opened 2019-11-08
Passing Error on Closed and Disconnected

### LINKED ISSUE #319: Add ConnectionErrorEventHandler
@ColinSullivan1 regarding implementing same/similar functionality as:
- https://github.com/nats-io/nats.go/pull/462/files
- https://github.com/nats-io/nats.go/pull/464

Is the idea that we add:

```csharp
public EventHandler<ConnErrorEventArgs> ConnectionErrorEventHandler = null;
```

and if that is hooked in, we use that upon disconnect due to error instead of `DisconnectedEventHandler`?

danielwertheim:
Resolves #319

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -1124,10 +1124,10 @@ namespace NATS.Client
                     }
                 }
             }
-            catch (Exception e)
+            catch (Exception ex)
             {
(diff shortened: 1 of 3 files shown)


ColinSullivan1 (approved):
LGTM - Thanks!

Issue #237, opened 2018-11-23
Authorization on _INBOX.> 


ramasa:
Please provide option to specify custom inbox prefix (say _INBOX.client1) , so that below authorization rule would prevent client from listening to other's response (_INBOX.>)

```
authorization {
    CLIENT1 = {
        publish = ["req.foo", "req.bar"]
        subscribe = "_INBOX.client1.>"
    }
```

Server is using java client and our customer using .NET 
More context here https://github.com/nats-io/java-nats/issues/192

sixlettervariables:
If I understand you correctly, you'd like to be able to change the inbox prefix per client; perhaps like so:
```csharp
var opts = ConnectionFactory.GetDefaultOptions();
opts.InboxPrefix = $"_INBOX.{clientId}.";
// ... other connection options
```

ramasa:
Yes, or by default use the nats username ( _INBOX.<nats_username> )?

ColinSullivan1:
Resolved by #322

PR #322, opened 2019-11-08
Custom inbox prefix

danielwertheim:
Adds support for custom inbox prefix to solve #237


danielwertheim:
After discussion I reverted from functional generators to simple string prefix.

Not sure if you think there should be validation when assigning the value? Are there any specific rules that should be enforced then?

Will complement with test.

ColinSullivan1:
Good point on validation....  You can validate with `Subscription.IsValidSubject(string subject)`

danielwertheim:
@ColinSullivan1 Validation added. Still lacks one or two integration tests.

danielwertheim:
@ColinSullivan1 @watfordgnf  I changed the validation as I do believe it can not be exactly as subjects. Might be wrong. Also added integration test which also found an issue. I've nothing else planned for this PR.

ColinSullivan1 (approved):
LGTM.  Thanks!

Issue #282, opened 2019-08-27
Handle expiration and auth errors on reconnects.


ColinSullivan1:
See https://github.com/nats-io/nats.go/pull/499

ColinSullivan1:
Resolved by #323

PR #323, opened 2019-11-12
Auth error reconnects

danielwertheim:
Solves #282 by adding support for retrying connect attempts upon auth failures.

The `Authentication expired` error was taken from documentation, I can not see that happen. As in, not sure how to test it. Tried with an expired JWT but get `Authentication Violation` for that.
@ColinSullivan1  Should it still be in?

Also, `Authentication Timeout` never reaches client as the server cuts the connection, so when we read from the stream, the underlying socket is disconnected. Maybe the server should have a short delay before cutting a client of?

@ColinSullivan1 Are there any other specific auth violations that we should look for?

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -1147,13 +1147,44 @@ namespace NATS.Client
             try
             {
                 exToThrow = null;
-                lock (mu)
+
+                NATSConnectionException natsAuthEx = null;
+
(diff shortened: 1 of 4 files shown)


ColinSullivan1:
...not sure about a server delay, that could exacerbate a DOS attack.  I think checking for both error conditions should be left in to maintain parity with the go client (and just in case it's a corner case we can reach testing locally).  The two errors should cover it.  Thanks!

ColinSullivan1 (approved):
LGTM.  Thanks!

Issue #299, opened 2019-10-09
IConnection.Request performs slower than manual request/reply


CrossBound:
So based on the comment here [https://github.com/nats-io/nats.net/issues/297#issuecomment-540189758](https://github.com/nats-io/nats.net/issues/297#issuecomment-540189758) I tried using the request api. Heres some interesting results. If I use RequestAsync like this: 
```csharp
        public Task GuaranteedDeliveryAsync(object Payload, CancellationToken CancellationToken = default)
        {
            if (_isDisposed)
            {
                throw new ObjectDisposedException(this.GetType().Name);
            }

            _counter++;
            return _connection.RequestAsync(_queue
                , JsonSerializer.SerializeToUtf8Bytes(new Payload<object>(_counter, Payload)), 150);
        }
```
I run about ~3,300+ msgs/sec with RTT of 1.12 ms /ea P99. My VS Analysis tools however show some interesting CPU and GC data.

![image](https://user-images.githubusercontent.com/11417093/66523192-e647f880-eab4-11e9-87e9-cd0465d79eb8.png)

However, if I change my code to look like this
```csharp
        public Task GuaranteedDeliveryAsync(object Payload, CancellationToken CancellationToken = default)
        {
            if (_isDisposed)
            {
                throw new ObjectDisposedException(this.GetType().Name);
            }

            _counter++;
            byte[] data = JsonSerializer.SerializeToUtf8Bytes(new Payload<object>(_counter, Payload));
            bool acknowledged = false;
            for (int i = 0; i < 3; i++)
            {
                _connection.Publish(_queue, _inbox, data);
                _connection.Flush();

                try
                {
                    while (true)
                    {
                        var response = _responseSubscription.NextMessage(100);
                        var ackID = BitConverter.ToUInt64(response.Data, 0);
                        if (ackID == _counter)
                        {
                            acknowledged = true;
                            break;
                        }
                    }
                }
                catch (NATSTimeoutException)
                {
                    // try again up to 3 times
                }

                if(acknowledged)
                {
                    break;
                }
            }

            if(acknowledged == false)
            {
                throw new Exception("Failed to successfully publish message");
            }

            return Task.CompletedTask;
        }
```

my performance improves significantly.... notice the number of messages per second as well as the GC collections and CPU data.

![image](https://user-images.githubusercontent.com/11417093/66523344-5a829c00-eab5-11e9-819f-4a4a6091e3f7.png)

CrossBound:
[...]

PR #324, opened 2019-11-12
Issue 299 mem leak

danielwertheim:
A first fix for #299 ensuring a token registration is cleaned up

Without this fix, MB consumption when monitoring in Visual Studio keeps going up. Will look further into that request construct but wanted to provide this fix first.


watfordgnf (approved):
Ack, I forgot to submit this fix. Looks good!

src/Samples/WinFormsSample/Form1.cs
```diff
@@ -0,0 +1,140 @@
...
+            _pubConnection?.Dispose();
+            _pubConnection = null;
+
+            _cts.Cancel();
+
+            _subConnection?.Dispose();
+            _subConnection = null;
+        }
+
+        private async void btnRun_Click(object sender, EventArgs e)
```
ColinSullivan1: Think it'd be a good idea to disable the button while running, or do you want to allow multiple parallel runs?

src/Samples/WinFormsSample/Form1.cs
```diff
@@ -0,0 +1,140 @@
...
+            {
+                var payload = Encoding.UTF8.GetBytes(Guid.NewGuid().ToString("N"));
+
+                var numOfMessages = numMessages.Value;
+                var configAwaitFalse = chkConfigureAwaitFalse.Checked;
+
+                for (var i = 0; i < numOfMessages; i++)
+                    await _pubConnection.RequestAsync(Subject, payload, _cts.Token).ConfigureAwait(!configAwaitFalse);
+            }));
+
```
ColinSullivan1, line 86: Small request - could you initialize one of the list items?  e.g.
`lstScenarios.SelectedItem = lstScenarios.Items[0];`

ColinSullivan1:
Edited:  Other than a few minor things and the windows form issue, LGTM.  Forms brings back some old memories.  :)

ColinSullivan1 (approved):

ColinSullivan1 (changes requested):
In other testing, it looks like the windows forms project prevents builds on linux and mac.  Could you make that project only build on the windows platform?

danielwertheim:
@ColinSullivan1 I've now fixed so that all projects can be built in a Linux Docker container and I hope that is good enough verification. Had to exclude explicit support for `.NET4.5` for **non-windows**. Is that OK?

It would be simpler if we stuck to .NET Standard and .NET Core app only but this works for now. To not build the Forms app I had to include a `LinuxRelease` and `LinuxDebug` build config do define that WinForms project should not be built then.

We could look at `LinuxDebug` and `LinuxRelease` instead of:

```
$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Windows)))
```

That way you control what to build in what env. Perhaps a cleaner approach? That would look like this: https://github.com/nats-io/nats.net/commit/4c0fd9a44319fc52620b846c554f9fc45b53e3b8

ColinSullivan1:
Thanks @danielwertheim.  

>  Had to exclude explicit support for .NET4.5 for non-windows. Is that OK?

Absolutely - the alternative is mono, which imo should be a custom build on a fork at this point.

I would like to keep .NET 4.5 for the time being, there are some users that still require that framework.

> We could look at LinuxDebug and LinuxRelease...

That sounds cleaner to me - might want to add a line in the readme about selecting Linux to building on OSx - we have some users that build .NET there.

ColinSullivan1 (approved):
LGTM - Thanks!

danielwertheim:
@ColinSullivan1  I'll add the config solution to this PR and complete/merge it when done.

ColinSullivan1:
Thanks!

PR #325, opened 2019-11-13
Disables creating package for WinForms sample

danielwertheim:
This slipped through in the last PR. Creating NuGet packages for samples is nothing we do.


ColinSullivan1 (approved):
LGTM.  Thanks.

Issue #315, opened 2019-11-03
Unable to fail over in clustered setup


CrossBound:
I brought this up in Slack but thought it might be better to just create an issue here so we can better track the issue. 

I have been testing a clustered setup and when one of my nodes goes offline my c# clients fail to connect to the other cluster nodes. Once the down node comes back online the client still will not reconnect and will have to be killed and restarted to get it to connect again. I've tried this with the latest pre-release and that made no difference. I've tried it on .net 4.7.2 as well as .net core 3 and both have the same result.

My last test did not include any TLS or auth and it still didn't work. I can connect to both servers directly. I'm clustering like this on the seed server:

`nats-server -DV -p 4222 -m 8222 --cluster nats://0.0.0.0:6222`

And on the secondary (172.16.1.118 is the seed server):  

`nats-server -p 4222 -m 8222 --cluster nats://0.0.0.0:6222 --routes nats://172.16.1.118:6222 -DV`

I've seen different exception messages depending on where the process was at the time the server went offline, but here is an example: 

```
NATS.Client.NATSException: Flush error. ---> System.IO.IOException: Unable to write data to the transport connection: An established connection was aborted by the software in your host machine. ---> System.Net.Sockets.SocketException: An established connection was aborted by the software in your host machine
  at System.Net.Sockets.NetworkStream.Write(Byte[] buffer, Int32 offset, Int32 size)
  --- End of inner exception stack trace ---
  at System.Net.Sockets.NetworkStream.Write(Byte[] buffer, Int32 offset, Int32 size)
  at System.Net.Security.SslStreamInternal.WriteSingleChunk[TWriteAdapter](TWriteAdapter writeAdapter, ReadOnlyMemory`1 buffer)
  at System.Net.Security.SslStreamInternal.WriteAsyncInternal[TWriteAdapter](TWriteAdapter writeAdapter, ReadOnlyMemory`1 buffer)
  at System.Net.Security.SslStreamInternal.Write(Byte[] buffer, Int32 offset, Int32 count)
  at System.Net.Security.SslStream.Write(Byte[] buffer, Int32 offset, Int32 count)
  at System.IO.BufferedStream.Flush()
  at NATS.Client.Connection.Flush(Int32 timeout)
  --- End of inner exception stack trace ---
  at NATS.Client.Connection.Flush(Int32 timeout)
[...]

PR #329, opened 2019-11-14
Handle zero length reads correctly

### LINKED ISSUE #326: Reconnect Hang
When the NATS server is recycled very quickly on the command line with ctrl+c, the client does not reconnect and seems to hang.

I was able to reproduce it with this code:

```csharp
            Options opts = ConnectionFactory.GetDefaultOptions();
            opts.AllowReconnect = true;
            opts.MaxReconnect = 100000;
            opts.ReconnectedEventHandler = (obj, args) =>
            {
                Console.WriteLine("reconnected");
            };
            opts.DisconnectedEventHandler = (obj, args) =>
            {
                Console.WriteLine("disconnected");
            };
            opts.ClosedEventHandler = (obj, args) =>
            {
                Console.WriteLine("closed");
            };
            opts.AsyncErrorEventHandler = (obj, args) =>
            {
                Console.WriteLine("async error: {0}", args.Error);
            };
            new ConnectionFactory().CreateConnection(opts);
            Thread.Sleep(100000);
```

ColinSullivan1:
The NATS server had an exit path on windows that would sometimes gracefully close a socket, which can create a zero length read on at NetworkStream.

The client assumed a zero length read would only occur when it closed the socket (not the server), so was keying on that to exit the readLoop.  When the server would gracefully shut sockets down, the client could encounter a zero length read on reconnect, killing the readLoop(), rendering the client unable to process messages.

This fix treats a zero length read as an error, and if we're in the process of reconnecting or closing, the processOpError method will do the right thing.

Many thanks to @danielwertheim for helping debug this problem.

Resolves #326, #315.

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -1776,10 +1776,17 @@ namespace NATS.Client
                 {
                     len = br.Read(buffer, 0, Defaults.defaultReadLength);
 
-                    // Socket has been closed gracefully by host
+                    // A length of zero can mean that the socket was closed
+                    // locally by the appliation (Close) or the server
+                    // gracefully closed the socket.  There are some cases
+                    // on windows where a server could take an exit path that
+                    // gracefully closed sockets.  Throw an exception so we
+                    // can reconnect.  If the network stream has been closed
+                    // by the client, processOpError will do the right thing
+                    // (nothing).
                     if (len == 0)
                     {
-                        break;
+                        throw new NATSConnectionException("Server closed the connection");
                     }
 
                     parser.parse(buffer, len);
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -1777,10 +1777,10 @@ namespace NATS.Client
                     len = br.Read(buffer, 0, Defaults.defaultReadLength);
 
                     // A length of zero can mean that the socket was closed
-                    // locally by the appliation (Close) or the server
+                    // locally by the application (Close) or the server
                     // gracefully closed the socket.  There are some cases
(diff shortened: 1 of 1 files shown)


watfordgnf (approved):

ColinSullivan1:
@watfordgnf, appreciate the review!

watfordgnf:
Tricky error condition.

ColinSullivan1:
Easy to fix, brutal to find.

src/NATS.Client/Conn.cs
```diff
@@ -1776,10 +1776,17 @@ private void readLoop()
...
+                    // gracefully closed the socket.  There are some cases
+                    // on windows where a server could take an exit path that
+                    // gracefully closes sockets.  Throw an exception so we
+                    // can reconnect.  If the network stream has been closed
+                    // by the client, processOpError will do the right thing
+                    // (nothing).
                     if (len == 0)
                     {
-                        break;
+                        throw new NATSConnectionException("Server closed the connection");
```
danielwertheim: Nitpicking maybe, but should the message end with a period? There seem to be different styles in use.

Also, not sure if we really need to throw and catch, or if we just could call and pass ex to `processOpError` as that assigns the exception to `lastEx` anyway. I don't see that we throw that. And there are similar constructs in other places:

```
lastEx = new NATSSlowConsumerException();
```

danielwertheim (approved):

src/NATS.Client/Conn.cs
ColinSullivan1: We could, although I kinda like the one code path to `processOpError`.  I'll fix the `.` and merge.

PR #331, opened 2019-11-18
Splits pipe into two, one for CI and one for CD

danielwertheim:
This so that we can simplify for forks and PRs from forks as the CI flow now does not require a NuGet connection. That is only required in the CD flow which is only needed from this repo.

Once this is in, I'll setup one more pipe in AZ for the CD flow.

### DIFF
--- a/az-templates/stage-build.yml
+++ b/az-templates/stage-build.yml
@@ -1,30 +1,3 @@
-name: $(Semver)
-
-variables:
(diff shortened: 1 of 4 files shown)


watfordgnf (approved):

danielwertheim:
@ColinSullivan1 I'll merge this one after we have released the other stuff.

PR #338, opened 2019-11-19
Uses packageIcon instead of packageIconUrl as url has been deprecated

danielwertheim:
Took the image and resized it to 64x64 ([result](https://github.com/danielwertheim/nats.net/blob/c38ac50a93872a234c35e0c80545cde139f5c7be/src/NATS.Client/package-icon.png)) as recommended in link below. Included it in the repo and then I use that when the package is created. We could perhaps download it from the CNCF repo but I do believe this is simpler and easier to understand.

More info:
- https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5048
- https://docs.microsoft.com/en-us/nuget/reference/nuspec#iconurl


watfordgnf (approved):
Looks good.

Issue #202, opened 2018-02-23
Performance of concurrent Request/Reply


plaisted:
I've was working on a basic RPC setup with NATS.Client and found large performance degradation using the built in Request feature if the connection is shared for multiple concurrent requests.

The basic code per concurrent request stream is essentially:
```csharp
        public static async Task Run(int count, IConnection connection, ManualResetEvent starter)
        {
            if (starter != null)
            {
                // ensure connection warm and wait
                await Run(10, connection, null);
                starter.WaitOne();
            }
            var sw = new Stopwatch();
            sw.Start();
            for (var i = 0; i < count; i++)
            {
                var result = await connection.RequestAsync("test", Encoding.UTF8.GetBytes("ping"));
            }
            sw.Stop();
            Console.WriteLine("Per request: " + Decimal.Divide(sw.ElapsedMilliseconds, count));
        }
```

If only one task is spawned I get an average round trip of under 1ms (0.6-0.7 typically).

If I spawn 10 tasks running, the average is approximately 10x higher (~7ms).

This makes it seem like the requests are unable to run concurrently at all. I would expect some performance degradation (they are sharing the same TCP connection after all) but a linear increase like this seems off. This suggests a single RPC is somehow saturating the connection (doesn't seem possible), or my code (or the library) is blocking in some inefficient way.

I've tested the same type of setup sharing the same network connection with no issues using GRPC or RabbitMQ. The performance dropoff was gradual and could get quite a large number concurrently.

Any suggestions on how to handle a concurrent requests like this? The final application is a web server publishing messages so the concurrency is driven by requests coming in.

ColinSullivan1:
@plaisted , thank you for raising this!  Your code looks straightforward to me; I'll take a look.  Which version are you running?

plaisted:
I installed the latest Nuget earlier today. I'll check later when I have access to the servers I was testing on. Thanks for quick response.

plaisted:
Running on Windows 10x64.
nats-server version 1.0.4
NATS.Client 0.8.0
.Net: netcoreapp2.0

ColinSullivan1:
Absolutely.  Thanks for the info!  I'll be out of my office over the weekend so this may take a few days, but I'll try to get to this ASAP.

plaisted:
Very simple async subscriber with a publish/flush mirroring an example. I
tried multiple subscriber apps on the same queue at the same time without
significant change in round trip time so appears to be on the request side.

[...]

PR #339, opened 2019-11-20
Request optimizations

danielwertheim:
After these changes when running the Benchmarks included I see an increase with about _20%_ in the request async flow.

When debugging and monitoring GC activity (as discussed in #299), both with and without a specified timeout, I see a highly reduced GC activity also leading to reducing CPU activity.

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -2812,14 +2812,13 @@ namespace NATS.Client
 
         private void requestResponseHandler(object sender, MsgHandlerEventArgs e)
         {
+            if (e.Message == null)
+                return;
+
             //               \
             //               \/
             //  _INBOX.<nuid>.<requestId>
-            string requestId = e.Message.Subject.Substring(globalRequestInbox.Length + 1);
-            if (e.Message == null)
-            {
-                return;
-            }
+            var requestId = e.Message.Subject.Substring(globalRequestInbox.Length + 1);
 
             bool isClosed;
             InFlightRequest request;
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -2665,10 +2665,14 @@ namespace NATS.Client
             request.Waiter.Task.ContinueWith(t => GC.KeepAlive(t.Exception), TaskContinuationOptions.OnlyOnFaulted);
             waitingRequests.TryAdd(request.Id, request);
 
+            if (globalRequestSubscription != null)
+                return request;
+
             lock (mu)
             {
-                if(globalRequestSubscription == null)
-                    globalRequestSubscription = subscribeAsync(string.Concat(globalRequestInbox, ".*"), null, requestResponseHandler);
+                if (globalRequestSubscription == null)
+                    globalRequestSubscription = subscribeAsync(string.Concat(globalRequestInbox, ".*"), null,
+                        requestResponseHandler);
             }
 
             return request;
@@ -2691,9 +2695,6 @@ namespace NATS.Client
             using (var request = setupRequest(timeout, CancellationToken.None))
             {
                 request.Token.ThrowIfCancellationRequested();
-
-                if (globalRequestSubscription == null)
-                    throw new NATSException("No global inbox subscription exists for receiving request responses in.");
 
                 publish(subject, string.Concat(globalRequestInbox, ".", request.Id), data, offset, count, true);
 
@@ -2734,9 +2735,6 @@ namespace NATS.Client
             using (var request = setupRequest(timeout, token))
             {
                 request.Token.ThrowIfCancellationRequested();
-
-                if (globalRequestSubscription == null)
-                    throw new NATSException("No global inbox subscription exists for receiving request responses in.");
 
                 publish(subject, string.Concat(globalRequestInbox, ".", request.Id), data, offset, count, true);
 
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -757,9 +757,6 @@ namespace NATS.Client
             return null;
         }
 
-        // Ensure we cannot instanciate a connection this way.
-        private Connection() { }
-
         /// <summary>
         /// Initializes a new instance of the <see cref="Connection"/> class
         /// with the specified <see cref="Options"/>.
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -162,68 +162,46 @@ namespace NATS.Client
         // used to map replies to requests from client (should lock)
         private long nextRequestId = 0;
 
-        private Dictionary<string, InFlightRequest> waitingRequests = 
+        private readonly Dictionary<string, InFlightRequest> waitingRequests = 
             new Dictionary<string, InFlightRequest>(StringComparer.OrdinalIgnoreCase);
 
         // Handles in-flight requests when using the new-style request/reply behavior
         private sealed class InFlightRequest : IDisposable
         {
-            public InFlightRequest(CancellationToken token, int timeout)
+            internal InFlightRequest(string id, CancellationToken token, int timeout, Action<string> onCompleted)
             {
+                this.Id = id;
                 this.Waiter = new TaskCompletionSource<Msg>();
-                if (token != CancellationToken.None)
+                this.onCompleted = onCompleted;
+                this.tokenSource = token == CancellationToken.None
+                    ? new CancellationTokenSource()
+                    : CancellationTokenSource.CreateLinkedTokenSource(token);
+                
+                this.tokenRegistration = this.tokenSource.Token.Register(() =>
                 {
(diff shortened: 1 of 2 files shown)


danielwertheim:
@ColinSullivan1 having the global subscription created immediately when connected, is that avoided due to not creating an unnecessary subscription? Could this be OK if we added config for the behavior? E.g `options.LazyGlobalSubscriptionInit = true`

danielwertheim:
Reverted a commit. As it's not possible to specify both `timeout` and `cancellationToken` in the public request API, I thought there was no need for a `CancellationTokenSource` when a Token had been specified but no timeOut had been specified. But it is. Otherwise it never times out and we get a deadlock. Will look into that further and potentially tweak in future PR.

src/Samples/WinFormsSample/Form1.cs
```diff
@@ -1,4 +1,6 @@
-﻿#if Windows
+﻿
+using System.Collections.Generic;
+#if Windows
```
watfordgnf: Should this using be within the `#if` block?

src/Samples/WinFormsSample/Form1.cs
```diff
@@ -146,20 +190,24 @@ private async void btnRun_Click(object sender, EventArgs e)
...
+                return;
+
             try
             {
-                var scenario = lstScenarios.SelectedItem as Scenario;
-                if (scenario == null)
-                    return;
+                await Task.Run(async () =>
+                {
+                    GC.Collect();
```
watfordgnf: Was this just for debugging?

src/NATS.Client/Conn.cs
```diff
@@ -2647,227 +2621,142 @@ public void Publish(string subject, string reply, byte[] data, int offset, int c
...
-                                throw;
-                            }
+                    return request.Waiter.Task.Result;
+                }
+                catch (AggregateException ae)
+                {
+                    foreach (var e in ae.Flatten().InnerExceptions)
+                    {
+                        // we *should* only have one, and it should be
+                        // a NATS timeout exception.
```
watfordgnf, line 2712: Could it also be a TaskCancelledException?

src/NATS.Client/Conn.cs
```diff
@@ -2647,227 +2621,142 @@ public void Publish(string subject, string reply, byte[] data, int offset, int c
...
-                    request.Id,
-                    request);
-            }
-
-            if (createSub)
-            {
-                globalRequestSubscription = subscribeAsync(globalRequestInbox + ".*", null, requestResponseHandler);
+            var requestId = Interlocked.Increment(ref nextRequestId);
+            if (requestId < 0) //Check if recycled
+                requestId = (requestId + long.MaxValue + 1);
```
watfordgnf, line 2662: Good addition, and likely one that won't impact the `ConcurrentDictionary` as apparently it can only hold `Int32.MaxValue` objects.

watfordgnf (approved):
Benchmarks on my machine are certainly faster. Looks good!

watfordgnf:
In my app's case, we have maybe only one or two request-reply usages. All the rest are publish. We use web services for request/reply (and authorization). In our case, lazily creating the global subscription would be our preference.

Is there a performance gain by creating it immediately for heavy request/reply scenarios?

src/Samples/WinFormsSample/Form1.cs
danielwertheim: Yes, it should. Auto import gone bad (Alt+Enter). Fixing

src/Samples/WinFormsSample/Form1.cs
danielwertheim: Yes and no. Depends on what we want. The form contains many scenarios and wanted to see behavior of GC per scenario without restarting. Could of course be an option. I'll remove and someone else can add it if needed.

src/NATS.Client/Conn.cs
danielwertheim: Do not know. Haven't changed that behavior as far as I know.

danielwertheim: Just tried it by cancelling a running request and the `OperationCanceledException` is not thrown on the sync path where this is. Added explicit handler for that ex to test.

src/NATS.Client/Conn.cs
danielwertheim: I'm not following. I'm guessing the dictionary held strings due to the fact that it's parsed out on the incoming reply from the subject and there you don't want to pay for converting the string to an int/long.

watfordgnf: You're correct , missed we used string rather than longs.

danielwertheim:
Possibly this will solve #202 as well

src/NATS.Client/Conn.cs
```diff
@@ -123,8 +123,6 @@ public Options Opts
         private Uri             url     = null;
         private ServerPool srvPool = new ServerPool();
 
-        private Dictionary<string, Uri> urls = new Dictionary<string, Uri>();
```
ColinSullivan1, line 126: Thanks - I wonder how much of this stuff I've left in....

src/NATS.Client/Conn.cs
ColinSullivan1: Yes, string was from the original implementation of reply subjects.  If we're using an `int`, I wonder if this would perform even better using ints in the dictionary and converting to a string when publishing.

src/NATS.Client/Conn.cs
```diff
@@ -2647,227 +2621,142 @@ public void Publish(string subject, string reply, byte[] data, int offset, int c
...
-
-        private InFlightRequest requestSync(string subject, byte[] data, int offset, int count, int timeout, CancellationToken token)
-        {
-            InFlightRequest request = setupRequest(timeout, token);
+            var request = new InFlightRequest(requestId.ToString(CultureInfo.InvariantCulture), token, timeout, removeOutstandingRequest);
+            request.Waiter.Task.ContinueWith(t => GC.KeepAlive(t.Exception), TaskContinuationOptions.OnlyOnFaulted);
+            waitingRequests.TryAdd(request.Id, request);
 
-            request.Token.ThrowIfCancellationRequested();
+            if (globalRequestSubscription != null)
```
ColinSullivan1, line 2668: It looks like this check should be under the lock below.

ColinSullivan1 (changes requested):
One question about int vs string for the request id, and there's one check that I think needs to be moved under a lock.  Other than that, LGTM.

src/NATS.Client/Conn.cs
danielwertheim: @watfordgnf resolving. Feel free to re-comment if the "manual test" above is not enough or if I misunderstood anything.

src/NATS.Client/Conn.cs
danielwertheim: @ColinSullivan1  I might be misunderstanding this, I mean, the lock statement will initialize the subscription once, right? So do we really need the lock statement for the check if it already exists?

src/NATS.Client/Conn.cs
danielwertheim: But wouldn't `requestResponseHandler` take a fair amount of perf hit if it first need to extract the string (as of today) then parse the string to an int (new addition if we change to ints) then do the lookup.

I do understand that memory will be reduced though (as long as > 9999 outstanding requests), but is that of any real concern in this case?

src/NATS.Client/Conn.cs
ColinSullivan1: Yeah...  it's possible the value could be cached and different threads would see different values.  Declaring the globalRequestSubscription to be `volatile` *might* work because there is locking below when the variable is used next (ensuring instruction ordering), but risky.  Since there's a locking involved anyhow, probably cleanest just to check under the lock.

This provides some context...
https://stackoverflow.com/questions/154551/volatile-vs-interlocked-vs-lock

src/NATS.Client/Conn.cs
ColinSullivan1: It could, good point.  Memory wouldn't be a big concern, and thinking about it, if there were performance increases here they'd be negligible in terms of the entire request/reply operation.  It's fine as is...

src/NATS.Client/Conn.cs
ColinSullivan1: After discussion on slack and reviewing this again, it's safe.  LGTM.

ColinSullivan1 (approved):

danielwertheim:
@ColinSullivan1  For ease, removing interlocked etc. a Guid would be easy. But I guess you don't want a longer subscription etc. than necessary

ColinSullivan1:
@danielwertheim , GUID would certainly work, but imo this will be faster and because the first token of the global inbox subject is a random set of bytes seeded by a GUI, we should be safe the way it is, and it'll reduce pressure at scale (smaller subjects as you mentioned).

PR #343, opened 2019-11-25
Improvements

danielwertheim:
- Improves flusher and read thread management:
  - On failure, try "clean" next thread
  - On success, remove from list of threads
- Fixes potential non cleaning of stream in Parser
- Fixes non disposal of Parser


src/NATS.Client/Conn.cs
```diff
@@ -923,16 +923,22 @@ private void waitForExits()
...
+            foreach (var t in cpy)
             {
                 try
                 {
-                    foreach (Thread t in wg)
-                    {
-                        t.Join();
-                    }
+                    t.Join();
+                    wg.Remove(t);
```
danielwertheim, line 936: @ColinSullivan1 @watfordgnf do we always want to remove the thread from the list or just on the success path? If always, we could just `Clear` the whole list afterwards.

watfordgnf: The two exception paths from Thread.Join are: (1) due to an invalid state (and would always throw), or (2) because the thread was interrupted.

I don't believe we interrupt those threads anywhere, so we should be okay to just clear the list at the end. That would also avoid the copy as well.

watfordgnf (approved):
Looks good.

src/NATS.Client/Conn.cs
ColinSullivan1: I agree, it should be OK to clear.

ColinSullivan1 (approved):
LGTM, if you want to add the clear go ahead.   Either way works for me.

src/NATS.Client/Conn.cs
danielwertheim: Fixed

danielwertheim: Unfixed. Reverted for now. Started to get lots of intermittent test failures. Tested in own pipeline as well.

PR #345, opened 2019-11-26
Ensures url has OK protocol.

### LINKED ISSUE #337: Clients hangs when trying to connect to a URL that is not NATS
Malformed URL throws exception but if you accidentally enter an URL that is valid but not nats-server the call hangs.

**Steps to reproduce:**

```
var cf = new ConnectionFactory();
var opts = ConnectionFactory.GetDefaultOptions()
opt.Servers = new[] {a valid URL to a web server e.g. http://localhost}
try
{
   var conn = cf.CreateConnection(opts)
}
catch(Exception ex)
{
   // Some Action
}
```
**Actual Result:**
Code hangs no exception thrown.
**Expected Result:**
Exception thrown

danielwertheim:
Closes #337 by checking for OK protocol.

### DIFF
--- a/src/NATS.Client/Options.cs
+++ b/src/NATS.Client/Options.cs
@@ -248,6 +248,8 @@ namespace NATS.Client
             }
         }
 
+        static readonly string[] protcolSep = new[] {"://"};
+        
         static string ensureProperUrl(string url)
         {
             if (string.IsNullOrWhiteSpace(url))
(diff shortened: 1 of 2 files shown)


src/NATS.Client/Options.cs
```diff
@@ -247,6 +248,24 @@ internal Options(Options o)
...
+            if (string.IsNullOrWhiteSpace(url))
+                return url;
+            
+            if (url.StartsWith("nats://", StringComparison.OrdinalIgnoreCase))
+                return url;
+
+            if (url.StartsWith("tls://", StringComparison.OrdinalIgnoreCase))
+                return url;
+
+            var parts = url.Split(new[]{"://"}, StringSplitOptions.RemoveEmptyEntries);
```
watfordgnf: Should hoist the allocation into a static.

watfordgnf (approved):
Small comment, otherwise looks good.

src/NATS.Client/Options.cs
danielwertheim: Fixed

ColinSullivan1 (approved):
LGTM - Thanks!

PR #347, opened 2019-11-27
Fixes so that conn.unsubscribe does not return null

danielwertheim:
Extracts CompletedTask and uses a completed task instead of null.


watfordgnf (approved):

src/NATS.Client/Internals/CompletedTask.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright 2015-2018 The NATS Authors
```
ColinSullivan1, line 1: Nitpick... 2019

ColinSullivan1 (approved):
LGTM.

PR #349, opened 2019-12-03
Fixes so that bw.Position is not used when NetworkStream

danielwertheim:
Only works when MemoryStream (pending) is used.

Case occurred here: https://github.com/nats-io/stan.net/issues/162

Tried to go through and find why we end up in this situation. Thought it was status access without lock but couldn't find it.

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -2496,7 +2496,7 @@ namespace NATS.Client
                         else
                             kickFlusher();
 
-                        if (bw.Position + count + pubProtoLen > rbsize)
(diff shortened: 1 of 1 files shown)


ColinSullivan1:
@danielwertheim, I think disabling the reconnect buffer would fix this, and would be in-line with the golang and C clients.  Given this is NATS streaming, I don't think the client should make an effort to resend - the streaming server will handle missed messages.

`opts.ReconnectBufSize = Options.ReconnectBufferDisabled`

ColinSullivan1 (approved):
LGTM.

PR #359, opened 2020-01-24
Fix #356:  Fix exceptions thrown by Drain

### LINKED ISSUE #356: Calling Drain on a closed connection throws NATSTimeoutException instead of NATSConnectionClosedException
Any exceptions that occur during the private drain(int timeout) method gets replaced by a NATSTimeoutException. This is particularly unexpected in the case that Drain is called on a closed connection:

https://github.com/nats-io/nats.net/blob/c50d7f6854a556fbc5d0bae646d147499dc683b9/src/NATS.Client/Conn.cs#L3897-L3939

ColinSullivan1:
Drain should throw a NATSConnectionClosedException when appropriate.

Resolves #356 

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -3890,6 +3890,13 @@ namespace NATS.Client
 
             lock (mu)
             {
+                if (isClosed())
+                    throw new NATSConnectionClosedException();
+
+                // if we're already draining, exit.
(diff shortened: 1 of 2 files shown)


watfordgnf (approved):
Looks good, couldn't get CI to come up to see if these passed.

ColinSullivan1:
Hmm...   You should be able to.  I'll double check perms.   
image

PR #365, opened 2020-02-12
Fix JetStream Reply Subjects

### LINKED ISSUE #364: Error when trying to pull message from jetstream
Trying out the new Jetstream server causes an issue with RequestAsync.

An exception is thrown when parsing the response

https://github.com/nats-io/nats.net/blob/3558232a673d65d4440050bb96492f2645f43892/src/NATS.Client/Conn.cs#L2597-L2605

Here, `e.Message.Subject = "foo.inc"` and `globalRequestInbox = "_INBOX.OpwL1p-dK8wVoYOZ20GWIy"`

```
Exception has occurred: CLR/System.ArgumentOutOfRangeException
Exception thrown: 'System.ArgumentOutOfRangeException' in System.Private.CoreLib.dll: 'startIndex cannot be larger than length of string.'
   at System.String.Substring(Int32 startIndex, Int32 length)
   at System.String.Substring(Int32 startIndex)
   at NATS.Client.Connection.requestResponseHandler(Object sender, MsgHandlerEventArgs e)
```

Adding a minimal example for the issue.

Running on osx and using latest nats-server and jsm utility from source but had the same issue running the `synadia/jsm:0.0.6` images in docker and `NATS.Client 0.10.0`

```cs

// # Start server
// ./nats-server -js

// test-jetstream % ./jsm str create
// ? Stream Name foo
// ? Subjects to consume foo.*
// ? Storage backend memory
// ? Retention Policy Limits
// ? Message count limit -1
// ? Message size limit -1
// ? Maximum message age limit -1
// ? Maximum individual message size -1
// Stream foo was created

// Information for Stream foo

// Configuration:

//              Subjects: foo.*
//      Acknowledgements: true
//             Retention: Memory - Limits
//              Replicas: 1
//      Maximum Messages: -1
//         Maximum Bytes: -1
//           Maximum Age: 0s
//  Maximum Message Size: -1
//     Maximum Consumers: -1

// State:

//             Messages: 0
//                Bytes: 0 B
//             FirstSeq: 0
//              LastSeq: 0
//     Active Consumers: 0



// test-jetstream % ./jsm con create
// ? Select a Stream foo
// ? Consumer name bar
// ? Delivery target
// ? Start policy (all, last, 1h, msg sequence) all
// ? Filter Stream by subject (blank for all)
// ? Maximum Allowed Deliveries -1
// Information for Consumer foo > bar

// Configuration:

//         Durable Name: bar
//            Pull Mode: true
//          Deliver All: true
//           Ack Policy: explicit
//             Ack Wait: 30s
//        Replay Policy: instant

// State:

//   Last Delivered Message: Consumer sequence: 0 Stream sequence: 0
//     Acknowledgment floor: Consumer sequence: 0 Stream sequence: 0
//         Pending Messages: 0
//     Redelivered Messages: 0


using System;
using System.Text;
using System.Threading.Tasks;
using NATS.Client;

namespace src
{
  public class Program
  {

    public static async Task Main(string[] args)
    {
      try
      {
        options.Url = "nats://localhost:4222";
        // Works when set to true
        // options.UseOldRequestStyle = true;

        using var connection = new ConnectionFactory().CreateConnection(options);

        // Publish some data to stream
        connection.Publish("foo.inc", Encoding.UTF8.GetBytes("Hello, World"));
        connection.Flush();

        // Try to read back one message
        var msg = await connection.RequestAsync("$JS.STREAM.foo.CONSUMER.bar.NEXT", Encoding.ASCII.GetBytes("1"));
        Console.WriteLine(Encoding.UTF8.GetString(msg.Data));
      }
      catch (Exception err)
      {
        Console.WriteLine(err.Message);
      }
    }
  }
}

```

```xml
<Project Sdk="Microsoft.NET.Sdk.Worker">
  <PropertyGroup>
    <TargetFramework>netcoreapp3.1</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <ProjectReference Include="..\nats.net\src\NATS.Client\NATS.Client.csproj" />
  </ItemGroup>
</Project>
```

ColinSullivan1:
JetStream will remap reply subjects for requests.  The request handling code needed to be updated to handle this.  This mirrors the go client's logic.

Resolves #364.

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -2592,47 +2592,46 @@ namespace NATS.Client
 
         protected Msg request(string subject, byte[] data, int offset, int count, int timeout) => requestSync(subject, data, offset, count, timeout);
 
-        private void removeOutstandingRequest(string requestId) => waitingRequests.TryRemove(requestId, out _);
+        private void RemoveOutstandingRequest(string requestId) => waitingRequests.TryRemove(requestId, out _);
 
-        private void requestResponseHandler(object sender, MsgHandlerEventArgs e)
+        private void RequestResponseHandler(object sender, MsgHandlerEventArgs e)
         {
             InFlightRequest request;
+            bool isClosed;
 
             if (e.Message == null)
                 return;
 
             var subject = e.Message.Subject;
 
-            // if it's a typical response, process normally.
-            if (subject.StartsWith(globalRequestInbox))
+            lock (mu)
             {
-                //               \
-                //               \/
-                //  _INBOX.<nuid>.<requestId>
-                var requestId = subject.Substring(globalRequestInbox.Length + 1);
-                if (!waitingRequests.TryGetValue(requestId, out request))
-                    return;
-            }
-            else
-            {
-                // We have a jetstream subject (remapped), so if there's only one
-                // request assume we're OK and handle it.
-                if (waitingRequests.Count == 1)
+                // if it's a typical response, process normally.
+                if (subject.StartsWith(globalRequestInbox))
                 {
-                    request = waitingRequests.ToArray()[0].Value;
+                    //               \
+                    //               \/
+                    //  _INBOX.<nuid>.<requestId>
+                    var requestId = subject.Substring(globalRequestInbox.Length + 1);
+                    if (!waitingRequests.TryGetValue(requestId, out request))
+                        return;
                 }
                 else
                 {
-                    // if we get here, we have multiple outsanding jetstream
-                    // requests.  We can't tell which is which we'll punt.
-                    return;
+                    // We have a jetstream subject (remapped), so if there's only one
+                    // request assume we're OK and handle it.
+                    if (waitingRequests.Count == 1)
+                    {
+                        request = waitingRequests.ToArray()[0].Value;
+                    }
+                    else
+                    {
+                        // if we get here, we have multiple outsanding jetstream
+                        // requests.  We can't tell which is which we'll punt.
+                        return;
+                    }
                 }
-            }
 
-            bool isClosed;
-
-            lock (mu)
-            {
                 isClosed = this.isClosed();
             }
 
@@ -2653,7 +2652,7 @@ namespace NATS.Client
             if (requestId < 0) //Check if recycled
                 requestId = (requestId + long.MaxValue + 1);
 
-            var request = new InFlightRequest(requestId.ToString(CultureInfo.InvariantCulture), token, timeout, removeOutstandingRequest);
+            var request = new InFlightRequest(requestId.ToString(CultureInfo.InvariantCulture), token, timeout, RemoveOutstandingRequest);
             request.Waiter.Task.ContinueWith(t => GC.KeepAlive(t.Exception), TaskContinuationOptions.OnlyOnFaulted);
             waitingRequests.TryAdd(request.Id, request);
 
@@ -2664,7 +2663,7 @@ namespace NATS.Client
             {
                 if (globalRequestSubscription == null)
                     globalRequestSubscription = subscribeAsync(string.Concat(globalRequestInbox, ".*"), null,
-                        requestResponseHandler);
+                        RequestResponseHandler);
             }
 
             return request;
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -155,8 +155,8 @@ namespace NATS.Client
         // used to map replies to requests from client (should lock)
         private long nextRequestId = 0;
 
-        private readonly ConcurrentDictionary<string, InFlightRequest> waitingRequests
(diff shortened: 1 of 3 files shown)


src/NATS.Client/Conn.cs
```diff
@@ -2596,15 +2596,38 @@ public void Publish(string subject, string reply, byte[] data, int offset, int c
...
+                //  _INBOX.<nuid>.<requestId>
+                var requestId = subject.Substring(globalRequestInbox.Length + 1);
+                if (!waitingRequests.TryGetValue(requestId, out request))
+                    return;
+            }
+            else
+            {
+                // We have a jetstream subject (remapped), so if there's only one
+                // request assume we're OK and handle it.
+                if (waitingRequests.Count == 1)
```
watfordgnf: We're not locked under `mu` here like in the go client's code, so I don't think this section is thread safe anymore:
https://github.com/nats-io/nats.go/blob/93a68d7e795f11c0aa30cfa38cf9a4702ae8d8b7/nats.go#L2630-L2670

I need to read up more on JetStream to have a better idea as to why only one request would be "out" with this model.

ColinSullivan1: Good catch, thanks.  I'll fix that.

With JetStream, you make a request to a "consumer" of a stream.  JetStream allows for the aggregation of multiple subjects (including wildcards), so it re-writes the subject (overriding the request's reply subject) to have the subject name of the original message.   A good place to start with JetStream right now is [here](https://github.com/nats-io/jetstream#readme).

watfordgnf: Reading that it appears we would need our consumer to tell us it is a JetStream consumer, so we could set it up properly in the waiting requests dictionary. Based on the golang code it seems that if you use a NATS connection for JetStream, you must ONLY use it for JetStream (given the map count logic). Is that correct?

ColinSullivan1: Sorry about that - JetStream is a standard NATS server with the JetStream subsystem enabled, so subscriptions on a connection can receive both persistent messages from a JetStream server side consumer and messages from standard producers.  On the same connection, one could make request to a service listening to "foo", in which case I'd get back a message on the inbox + request id reply subject originally used.    One could also make a request to a JetStream consumer (we used to call it observable), which would send the client a message using the original subject as the subject.  So if I've setup JetStream to aggregate subjects "foo" and "bar" under a "baz" JetStream server consumer, I'd make a request to "baz" and the subject could be overridden by the NATS server to be "foo", to provide the original subject, instead of the original reply subject we gave the message.

watfordgnf: If we're using async request-reply, why couldn't we have a situation where we'd have a pending request from a non-JS source and a JS sourced request at the same time? e.g. `waitingRequests.Count > 1`

ColinSullivan1: We could and so just make a best effort attempt there.  Since we've lost information in the system (the reply subject of the original message) we cannot accurately determine which received message should correlate to which request we're tracking.  It's not ideal, but this lets the application retry. 
 Hmm...   this situation would create a memory leak with ghost jetstream requests.

I suppose in this case we could search the outstanding requests to check if there's only one jetstream request to be a little better about this.

@kozlovic wdyt?

src/NATS.Client/Conn.cs
```diff
@@ -2592,47 +2592,46 @@ public void Publish(string subject, string reply, byte[] data, int offset, int c
 
         protected Msg request(string subject, byte[] data, int offset, int count, int timeout) => requestSync(subject, data, offset, count, timeout);
 
-        private void removeOutstandingRequest(string requestId) => waitingRequests.TryRemove(requestId, out _);
+        private void RemoveOutstandingRequest(string requestId) => waitingRequests.TryRemove(requestId, out _);
```
watfordgnf: This would have to be done while holding `mu` as well as the addition to `waitingRequests` in `setupRequest`. At that point it wouldn't be advantageous to use a `ConcurrentDictionary` as we'd be handling its concurrency.

ColinSullivan1: I agree, and ConcurrentDictionary is causing problems in Unity right now.  I'll give it a shot.

src/NATS.Client/Conn.cs
kozlovic: I agree that the code (even in Go) is confusing. There is too many assumptions and even if you don't mix non-JS and JS requests, as long as you have more than one request it would not work, which is not reasonable. I think this needs to be discussed and revisited.

ColinSullivan1: In that case let's keep the current code and if go is updated we'll make appropriate changes to the other clients (including this one).  Thanks @kozlovic.

ColinSullivan1: @watfordgnf, are we good to go?

kozlovic (approved):
LGTM. (Granted it may not be the correct approach, but that's what the Go client is doing at the moment. May need to be revisited later).

PR #366, opened 2020-02-23
Make exceptions constructors public to enable mocking

kaspertygesen:
I would like to be able to mock exceptions thrown by NATS.

NSubstitute example:
```
nats.Request(payload).ThrowsForAnyArgs(new NATSTimeoutException());
```

It would be great to be able to easily test for different NATS exceptions using mocks like this.


ColinSullivan1 (approved):
LGTM - Thanks for the contribution!

PR #374, opened 2020-04-28
Add ClientIP Connection property

ColinSullivan1:
Add the ClientIP connection property to retrieve the server info client IP field.

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -26,6 +26,7 @@ using System.Globalization;
 using System.Diagnostics;
 using NATS.Client.Extensions;
 using NATS.Client.Internals;
+using System.Net;
 
 namespace NATS.Client
 {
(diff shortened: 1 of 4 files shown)


kozlovic (approved):
LGTM

src/NATS.Client/Conn.cs
```diff
@@ -987,6 +988,27 @@ public string ConnectedUrl
...
+        {
+            get
+            {
+                lock (mu)
+                {
+                    if (string.IsNullOrEmpty(info.client_ip))
+                        return null;
+
+                    return IPAddress.Parse(info.client_ip);
+                }
```
watfordgnf: We can avoid locking during the IP address parsing:
```suggestion
                string clientIp;
                lock (mu)
                {
                    clientIp = info.client_ip;
                }
                
                return !String.IsNullOrEmpty(clientIp) ? IPAddress.Parse(clientIp) : null;
```

src/Tests/IntegrationTests/TestConnection.cs
```diff
@@ -571,8 +571,17 @@ public void CanConnectWhenHandshakeTimeoutIsSpecified()
             }
         }
 
-        /// NOT IMPLEMENTED:
-        /// TestErrOnMaxPayloadLimit
+        [Fact]
```
watfordgnf, line 574: A category or other attribute describing the minimum NATS version for some of these feature tests would be nice:
```suggestion
        [Fact]
        [Trait("Category", "NATS 2.1.6+")]
```

src/NATS.Client/Conn.cs
```diff
@@ -26,6 +26,7 @@
 using System.Diagnostics;
 using NATS.Client.Extensions;
 using NATS.Client.Internals;
+using System.Net;
```
watfordgnf: `using`s  got out of order.

watfordgnf (approved):
One comment and two suggestions, but otherwise looks good!

src/NATS.Client/Conn.cs
ColinSullivan1: Thanks - good catch.

kozlovic (approved):
LGTM

watfordgnf (approved):

ColinSullivan1:
Thanks for the reviews!  Once tests are green, I'll merge.

PR #376, opened 2020-05-06
Randomize implicit servers

ColinSullivan1:
See https://github.com/nats-io/nats.go/issues/565.

The server list is randomized on initial setup, then implicit servers are added into the list in a random position.  The NoRandomize option is honored, which will keep the initial server list static and place implicit servers at the end of the server list.

This PR also includes unit testing for the server pool.  

### DIFF
--- a/src/NATS.Client/ServerPool.cs
+++ b/src/NATS.Client/ServerPool.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2016-2018 The NATS Authors
+﻿// Copyright 2016-2020 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


src/Tests/UnitTests/TestServerPool.cs
```diff
@@ -0,0 +1,139 @@
+﻿// Copyright 2015-2020 The NATS Authors
```
kozlovic: Since it is new file, start at 2020

kozlovic (approved):
LGTM, just a note about the copyright date for the new test file.

watfordgnf (approved):

src/Tests/UnitTests/TestServerPool.cs
ColinSullivan1: Will fix...  Thanks!

src/NATS.Client/ServerPool.cs
```diff
@@ -200,21 +203,30 @@ internal string[] GetServerList(bool implicitOnly)
...
             {
                 if (sList.Contains(s, duplicateSrvCheck))
                     return false;
 
-                sList.AddLast(s);
+                if (s.isImplicit && randomize)
+                {
+                    // pick a random spot to add the server.
+                    var randElem = sList.ElementAt(rand.Next(sList.Count));
+                    sList.AddAfter(sList.Find(randElem), s);
```
kozlovic, line 224: Based on latest comment on the Go client, I wonder if this need to be updated. Basically in Go I was shuffling the pool after adding all new URLs (when processing INFO). But the point made was that possibly the first URL in the pool would be moved (say to position 2 (index 1)), and on reconnect that would be the very next one to be tried although we just disconnected from it. Let me know if I make sense (note the Go PR has not been updated as I type this).

ColinSullivan1: Per discussions in Slack, we've concluded that this achieves the same result, so I'll keep this as is.  Appreciate the catch here!

watfordgnf (approved):

PR #379, opened 2020-05-15
Increase internal buffer sizes

### LINKED ISSUE #377: Publish  through connect.ngs.global
Hi,
I want to try to publish through connect.ngs.global but cannot get any message and i don't have any error.
Are the options I made were enough?

```csharp
            options.Url = "connect.ngs.global";
            options.SetUserCredentials(@"S:\syndia-nats\synadia-keys\.nkeys\creds\synadia\NATS_CONNECT\NATS_CONNECT.creds");
            options.Secure = true;
```

Thank you

ColinSullivan1:
The client was having trouble connecting to servers returning large info JSON payloads at connect time.  Unfortunately, this issue was surfacing as a TLS error, so took awhile to track down.

Resolves #377 

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -737,7 +737,7 @@ namespace NATS.Client
             CRLF_BYTES_LEN = CRLF_BYTES.Length;
 
             // predefine the start of the publish protocol message.
-            buildPublishProtocolBuffer(Defaults.MaxControlLineSize);
+            buildPublishProtocolBuffer(512);
 
             callbackScheduler.Start();
 
@@ -1381,6 +1381,7 @@ namespace NATS.Client
             StreamReader sr = null;
             try
             {
+                // TODO:  Make this reader (or future equivalent) unbounded.
                 // we need the underlying stream, so leave it open.
                 sr = new StreamReader(br, Encoding.UTF8, false, Defaults.MaxControlLineSize, true);
                 result = sr.ReadLine();
@@ -1813,24 +1814,6 @@ namespace NATS.Client
         // Roll our own fast conversion - we know it's the right
         // encoding. 
         char[] convertToStrBuf = new char[Defaults.MaxControlLineSize];
-
-        // Caller must ensure thread safety.
-        private string convertToString(byte[] buffer, long length)
-        {
-            // expand if necessary
-            if (length > convertToStrBuf.Length)
-            {
-                convertToStrBuf = new char[length];
-            }
-
-            for (int i = 0; i < length; i++)
-            {
-                convertToStrBuf[i] = (char)buffer[i];
-            }
-
-            // This is the copy operation for msg arg strings.
-            return new string(convertToStrBuf, 0, (int)length);
-        }
 
         // Since we know we don't need to decode the protocol string,
         // just copy the chars into bytes.  This increased
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -737,7 +737,7 @@ namespace NATS.Client
             CRLF_BYTES_LEN = CRLF_BYTES.Length;
 
             // predefine the start of the publish protocol message.
-            buildPublishProtocolBuffer(Defaults.scratchSize);
+            buildPublishProtocolBuffer(Defaults.MaxControlLineSize);
 
             callbackScheduler.Start();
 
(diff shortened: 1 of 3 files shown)


src/NATS.Client/Conn.cs
```diff
@@ -737,7 +737,7 @@ internal Connection(Options options)
             CRLF_BYTES_LEN = CRLF_BYTES.Length;
 
             // predefine the start of the publish protocol message.
-            buildPublishProtocolBuffer(Defaults.scratchSize);
+            buildPublishProtocolBuffer(Defaults.MaxControlLineSize);
```
kozlovic: Not sure what that does, but does not look right to be. If this deals with a publish message, it should not be limited to the control line. If it is just the "PUB" header of the PUB protocol, then scratchSize was probably enough.

src/NATS.Client/Conn.cs
```diff
@@ -1382,7 +1382,7 @@ private void sendConnect()
             try
             {
                 // we need the underlying stream, so leave it open.
-                sr = new StreamReader(br, Encoding.UTF8, false, 512, true);
+                sr = new StreamReader(br, Encoding.UTF8, false, Defaults.MaxControlLineSize, true);
```
kozlovic, line 1386: You should not limit the CONNECT to that. The connect should be as long as it needs to be.

src/NATS.Client/Conn.cs
```diff
@@ -1432,7 +1432,7 @@ private Control readOp()
             // the string directly using the buffered reader.
             //
             // Keep the underlying stream open.
-            using (StreamReader sr = new StreamReader(br, Encoding.ASCII, false, 1024, true))
+            using (StreamReader sr = new StreamReader(br, Encoding.ASCII, false, Defaults.MaxControlLineSize, true))
```
kozlovic, line 1436: Probably the place where that could have been an issue. You need to be able to read as much of the INFO is.

src/NATS.Client/Conn.cs
```diff
@@ -1812,7 +1812,7 @@ internal void deliverMsgs(Channel<Msg> ch)
 
         // Roll our own fast conversion - we know it's the right
         // encoding. 
-        char[] convertToStrBuf = new char[Defaults.scratchSize];
+        char[] convertToStrBuf = new char[Defaults.MaxControlLineSize];
```
kozlovic, line 1816: Again here, not sure about what the limit applies to.

kozlovic (approved):
Need to check the use of the limit on several places.

src/NATS.Client/Conn.cs
watfordgnf: I believe that argument specifies the *minimum* buffer size. The .NET default for streams is 1024 bytes and 4096 bytes if it is a file stream.

src/NATS.Client/Conn.cs
watfordgnf: This is also a case where the size is simply the minimum.

watfordgnf:
These places with expanded buffers already expanded them as needed. I'm wondering why increasing the default *minimum* buffer sizes fixed this issue specifically?

src/NATS.Client/Conn.cs
ColinSullivan1: Will fix, reset back to 512.

src/NATS.Client/Conn.cs
ColinSullivan1: I agree, but will leave this as a TODO today.  The only constructor that allows the stream to remain open requires a buffer size (which is a minimum), so would require a rewrite of this section.

src/NATS.Client/Conn.cs
ColinSullivan1: (Ivan) I agree, but see the comment above.

ColinSullivan1:
I believe is was the scratch size that fixed this, but wanted to "standardize" on default buffer sizes.

src/NATS.Client/Conn.cs
ColinSullivan1: This is used to avoid memory allocations while extracting message args...  MaxControlLine seemed appropriate imo.

src/NATS.Client/Conn.cs
kozlovic: But if the INFO is currently around ~520 bytes, it seem that it would not have been the issue.

kozlovic (approved):
LGTM.. since changing those sizes fixed the issue.

ColinSullivan1:
Ivan and I had a discussion, debugged, and I suspect the stream reader atop the buffered reader isn't properly moving the buffered reader's position, stopping at it's minimum buffer size.  It's a guess.  However, increasing this minimum buffer size eliminates the problem consistently (behaving as it did before INFO messages were > 512b).  Longer term, we should a)  dynamically allocate protocol messages buffers, and b) eliminate the stream readers to implement our own methods to read the buffered reader stream byte by byte up until the \r\n.

PR #385, opened 2020-05-29
Add Reconnect Jitter

ColinSullivan1:
Add some jitter to the reconnect wait.

See:
 - https://github.com/nats-io/nats.go/issues/563
 - https://github.com/nats-io/nats.go/pull/564

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -149,7 +149,7 @@ namespace NATS.Client
         Timer               ptmr = null;
         int                 pout = 0;
 
-        internal AutoResetEvent ReconnectDelayARE = new AutoResetEvent(false);
+        internal static Random random = new Random();
 
         private AsyncSubscription globalRequestSubscription;
         private readonly string globalRequestInbox;
@@ -1533,11 +1533,9 @@ namespace NATS.Client
 
         private void DefaultReconnectDelayHandler(object o, ReconnectDelayEventArgs args)
         {
-            Random rand = new Random();
-            int jitter = srvPool.HasSecureServer() ? rand.Next(opts.ReconnectJitterTLS) : rand.Next(opts.ReconnectJitter);
+            int jitter = srvPool.HasSecureServer() ? random.Next(opts.ReconnectJitterTLS) : random.Next(opts.ReconnectJitter);
 
-            ReconnectDelayARE.Reset();
-            ReconnectDelayARE.WaitOne(opts.ReconnectWait + jitter);
+            Thread.Sleep(opts.ReconnectWait + jitter);
         }
 
         // Try to reconnect using the option parameters.
@@ -3715,8 +3713,6 @@ namespace NATS.Client
 
             lock (mu)
             {
-                ReconnectDelayARE?.Set();
-
                 // Clear any queued pongs, e.g. pending flush calls.
                 clearPendingFlushCalls();
                 if (pending != null)
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -147,8 +147,9 @@ namespace NATS.Client
         internal Exception lastEx;
 
         Timer               ptmr = null;
-
         int                 pout = 0;
+
+        internal AutoResetEvent ReconnectDelayARE = new AutoResetEvent(false);
 
         private AsyncSubscription globalRequestSubscription;
         private readonly string globalRequestInbox;
@@ -723,6 +724,10 @@ namespace NATS.Client
         internal Connection(Options options)
         {
             opts = new Options(options);
+            if (opts.ReconnectDelayHandler == null)
+            {
+                opts.ReconnectDelayHandler = DefaultReconnectDelayHandler;
+            }
 
             PING_P_BYTES = Encoding.UTF8.GetBytes(IC.pingProto);
             PING_P_BYTES_LEN = PING_P_BYTES.Length;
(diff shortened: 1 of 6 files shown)


kozlovic (approved):
LGTM about the logic, will let others judge on the .NET language specifics.

src/NATS.Client/Conn.cs
```diff
@@ -1526,6 +1531,15 @@ private void scheduleConnEvent(EventHandler<ConnEventArgs> connEvent, Exception
             }
         }
 
+        private void DefaultReconnectDelayHandler(object o, ReconnectDelayEventArgs args)
+        {
+            Random rand = new Random();
```
watfordgnf: This should be a static instance.

src/NATS.Client/Conn.cs
```diff
@@ -1562,43 +1576,43 @@ private void doReconnect()
 
                 scheduleConnEvent(Opts.DisconnectedEventHandler, errorForHandler);
 
-                // TODO:  Look at using a predicate delegate in the server pool to
-                // pass a method to, but locking is complex and would need to be
-                // reworked.
                 Srv cur;
+                int wlf = 0;
```
watfordgnf, line 1577: `wlf`?

src/NATS.Client/NATS.cs
```diff
@@ -186,6 +196,29 @@ internal ConnEventArgs(Connection c, Exception error = null)
...
+    {
+        internal ReconnectDelayEventArgs(int attempts)
+        {
+            Attempts = attempts;
+        }
+
+        /// <Summary>
+        /// Gets the number of times the client has traversed the
+        /// server list in attempting to reconnect.
+        /// </Summary>>
```
watfordgnf:
```suggestion
        /// </Summary>
```

src/NATS.Client/NATS.cs
```diff
@@ -186,6 +196,29 @@ internal ConnEventArgs(Connection c, Exception error = null)
         public Exception Error { get; }
     }
 
+    /// <summary>
+    /// Provides details for the ReconnectDelayEvent.
+    /// </summary>
+    /// <remarks>
+    /// This event handler is a good place to apply backoff logic.  The associated
+    /// connection will be RECONNECTING so access or calling methods will result
```
watfordgnf:
```suggestion
    /// connection will be RECONNECTING so accessing or calling IConnection methods will result
```

src/NATS.Client/NATS.cs
```diff
@@ -186,6 +196,29 @@ internal ConnEventArgs(Connection c, Exception error = null)
...
+    /// in undefined behavior.
```
watfordgnf:
```suggestion
    /// in undefined behavior (including deadlocks).
```

src/NATS.Client/NATS.cs
```diff
@@ -186,6 +196,29 @@ internal ConnEventArgs(Connection c, Exception error = null)
         public Exception Error { get; }
     }
 
+    /// <summary>
+    /// Provides details for the ReconnectDelayEvent.
+    /// </summary>
+    /// <remarks>
+    /// This event handler is a good place to apply backoff logic.  The associated
```
watfordgnf, line 203: We may want to mention that assigning a non-default value to this event handler means the user is responsible for reconnecting backoff.

src/NATS.Client/Conn.cs
```diff
@@ -147,9 +147,10 @@ public Options Opts
         internal Exception lastEx;
 
         Timer               ptmr = null;
-
         int                 pout = 0;
 
+        internal AutoResetEvent ReconnectDelayARE = new AutoResetEvent(false);
```
watfordgnf: This should get disposed in the `Dispose` method.

watfordgnf (commented):
Some questions and comments.

src/NATS.Client/Conn.cs
watfordgnf: Also, why an `ARE` instead of just `Thread.Sleep`?

src/Tests/IntegrationTests/TestReconnect.cs
```diff
@@ -568,6 +558,92 @@ public void TestReconnectBufferBoundary()
...
+                    // Wait a bit for the reconnect loop to go into wait mode.
+                    Thread.Sleep(250);
+
+                    // Close the connection to break waiting
+                    c.Close();
+                    sw.Stop();
+
+                    // It should be around 400 ms max, but only need to check that
+                    // it's less than 30000.
+                    Assert.True(sw.ElapsedMilliseconds < 2000);
```
watfordgnf: 30 seconds or 2 seconds?

src/NATS.Client/Conn.cs
ColinSullivan1: Will fix - thanks!

src/NATS.Client/Conn.cs
ColinSullivan1: Copied from go for consistency / portability.  It stands for "whole list failed".

src/Tests/IntegrationTests/TestReconnect.cs
ColinSullivan1: Will fix...

src/NATS.Client/Conn.cs
ColinSullivan1: It seemed cleaner in that author intent is expressed, but I'll test with a Sleep to see if it gets interrupted.

src/NATS.Client/NATS.cs
ColinSullivan1: Will do, thanks!

src/NATS.Client/Conn.cs
ColinSullivan1: Thread.Sleep() worked....  Simpler and easier.  Thanks!

ColinSullivan1:
Works locally, but need to fix some flappers in CI.

PR #394, opened 2020-07-10
Add Message Headers

ColinSullivan1:
Adding NATS message headers to parallel https://github.com/nats-io/nats.go/pull/560.

Also see https://github.com/nats-io/nats-server/blob/master/doc/adr/0004-nats-headers.md

This also includes a change in CI testing to pull the NATS server from a dependency repository.  This allows for testing against unreleased edge features.

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -170,6 +170,9 @@ namespace NATS.Client
         private byte[] PUB_P_BYTES = null;
         private int    PUB_P_BYTES_LEN = 0;
 
+        private byte[] HPUB_P_BYTES = null;
+        private int    HPUB_P_BYTES_LEN = 0;
+
         private byte[] CRLF_BYTES = null;
         private int    CRLF_BYTES_LEN = 0;
 
@@ -738,21 +741,24 @@ namespace NATS.Client
             PUB_P_BYTES = Encoding.UTF8.GetBytes(IC._PUB_P_);
             PUB_P_BYTES_LEN = PUB_P_BYTES.Length;
 
+            HPUB_P_BYTES = Encoding.UTF8.GetBytes(IC._HPUB_P_);
+            HPUB_P_BYTES_LEN = HPUB_P_BYTES.Length;
+
             CRLF_BYTES = Encoding.UTF8.GetBytes(IC._CRLF_);
             CRLF_BYTES_LEN = CRLF_BYTES.Length;
 
             // predefine the start of the publish protocol message.
-            buildPublishProtocolBuffer(512);
+            buildPublishProtocolBuffers(512);
 
             callbackScheduler.Start();
 
             globalRequestInbox = NewInbox();
         }
 
-        private void buildPublishProtocolBuffer(int size)
+        private void buildPublishProtocolBuffers(int size)
         {
             pubProtoBuf = new byte[size];
-            Buffer.BlockCopy(PUB_P_BYTES, 0, pubProtoBuf, 0, PUB_P_BYTES_LEN);
+            Buffer.BlockCopy(HPUB_P_BYTES, 0, pubProtoBuf, 0, HPUB_P_BYTES_LEN);
         }
 
         // Ensures that pubProtoBuf is appropriately sized for the given
@@ -762,14 +768,14 @@ namespace NATS.Client
         {
             // Publish protocol buffer sizing:
             //
-            // PUB_P_BYTES_LEN (includes trailing space)
+            // HPUB_P_BYTES_LEN (includes trailing space)
             //  + SUBJECT field length
             //  + SIZE field maximum + 1 (= log(2147483647) + 1 = 11)
             //  + (optional) REPLY field length + 1
 
-            int pubProtoBufSize = PUB_P_BYTES_LEN
+            int pubProtoBufSize = HPUB_P_BYTES_LEN
                                 + (1 + subject.Length)
-                                + (11)
+                                + (34) // Include payload/header sizes and stamp
                                 + (reply != null ? reply.Length + 1 : 0);
 
             // only resize if we're increasing the buffer...
@@ -784,7 +790,7 @@ namespace NATS.Client
                 pubProtoBufSize |= pubProtoBufSize >> 16;
                 pubProtoBufSize++;
 
-                buildPublishProtocolBuffer(pubProtoBufSize);
+                buildPublishProtocolBuffers(pubProtoBufSize);
             }
         }
 
@@ -2003,7 +2009,7 @@ namespace NATS.Client
             int i = 0;
 
             // We only support 4 elements in this protocol version
-            for ( ; i < length && count < 4; i++)
+            for ( ; i < length && count < 5; i++)
             {
                 convertToStrBuf[i] = (char)buffer[i];
                 if (buffer[i] == ' ')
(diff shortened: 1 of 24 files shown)


src/NATS.Client/Info.cs
```diff
@@ -54,10 +56,11 @@ public static ServerInfo CreateFromJson(string json)
...
                 tls_required = x["tls_required"].AsBool,
-                max_payload =  x["max_payload"].AsLong,
-                connect_urls =  x["connect_urls"].Children.Select(n => n.Value).ToArray(),
-                nonce =  x["nonce"].Value,
-                proto = x["proto"].AsInt
+                max_payload = x["max_payload"].AsLong,
+                connect_urls = x["connect_urls"].Children.Select(n => n.Value).ToArray(),
+                nonce = x["nonce"].Value,
+                proto = x["proto"].AsInt,
+                headers = x["headers"].AsBool
```
watfordgnf:
```suggestion
                headers = x["headers"].AsBool,
```

src/NATS.Client/Info.cs
```diff
@@ -125,7 +131,8 @@ internal StringBuilder AppendAsJsonTo(StringBuilder sb)
                 ["jwt"] = jwt ?? string.Empty,
                 ["nkey"] = nkey ?? string.Empty,
                 ["sig"] = sig ?? string.Empty,
-                ["echo"] = echo
+                ["echo"] = echo,
+                ["headers"] = headers
```
watfordgnf: To avoid diffs including this line when another property is added:
```suggestion
                ["headers"] = headers,
```

src/NATS.Client/Msg.cs
```diff
@@ -13,13 +13,178 @@
...
+
+        // Message headers are in the form of:
+        // |HEADER|crlf|key1:value1|crlf|key2:value2|crlf|...|crlf
+        // e.g. MATS/1.0\r\nkey1:value1\r\nkey2:value2\r\n\r\n
+
+        // Define message header version string and size.
+        internal static readonly string Header = "NATS/1.0\r\n";
+        internal static readonly byte[] HeaderBytes = Encoding.UTF8.GetBytes(Header);
+        internal static readonly int    HeaderLen = HeaderBytes.Length;
+        internal static readonly int    MinimalValidHeaderLen = Encoding.UTF8.GetBytes(Header + "k:v\r\n\r\n").Length;
```
watfordgnf: Does this mean there is always a value?

src/NATS.Client/Msg.cs
```diff
@@ -13,13 +13,178 @@
...
+                {
+                    throw new NATSInvalidHeaderException();
+                }
+            }
+
+            // Remove the header identifier and trailing crlfs when creating the string
+            string kvs = Encoding.UTF8.GetString(bytes, HeaderLen, byteCount - HeaderLen - 4);
+
+            // Split Name Value Pairs
+            string[] kvpairs = kvs.Split(crlf, StringSplitOptions.RemoveEmptyEntries);
```
watfordgnf: These two splits will start to get expensive on the fast path due to the sheer volume of allocations, likely need to include a TODO to avoid them in favor of getting a UTF8 string directly from the underlying buffer using offsets.

src/NATS.Client/Msg.cs
```diff
@@ -13,13 +13,178 @@
...
+                Add(nvpair[0], nvpair[1]);
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the string entry with the specified string key in the message header.
+        /// </summary>
+        /// <param name="name">The string key of the entry to locate. The key can be null.</param>
+        /// <returns>A string that contains the comma-separated list of values associated with the specified key, if found; otherwise, null</returns>
+        new public string this[string name]
```
watfordgnf: As this is a public class conceptually someone could use it as a `NameValueCollection` and then the serialization trigger would never execute due to `new`. This would break the behavior:
```csharp
void AddHeaders(NameValueCollection c)
{
    c["Content-Type"] = "application/json";
}
```

src/NATS.Client/Parser.cs
```diff
@@ -21,6 +21,7 @@ internal sealed class MsgArg
 	    internal string subject;
 	    internal string reply;
 	    internal long   sid;
+        internal int    hdr;
```
watfordgnf, line 24:
```suggestion
          internal int    hdr;
```

watfordgnf (commented):
A few suggestions and comments.

src/NATS.Client/Msg.cs
ColinSullivan1: Good point... while there's some debate it looks like some HTTP header implementations will support empty values - we should as well.  I'll update and test for that.  Thanks.

src/NATS.Client/Msg.cs
ColinSullivan1: Agreed...  Will do.

src/NATS.Client/Msg.cs
ColinSullivan1: Do you think encapsulation would be a better pattern to use here?

watfordgnf: Yeah, given we don't have virtual access to the setter.

ColinSullivan1:
I think this should be good to go, would appreciate another quick pass @watfordgnf.

I also addressed the performance issues with Splits.  Had to move up to .NET 4.6 to use the `NameValueCollection` class - while documentation says it's supported in 4.5, the assembly [requires 4.6](https://www.nuget.org/packages/System.Collections.Specialized/4.3.0#dependencies).  

IMO it should be OK to move to 4.6 as the oldest supported framework version.

src/NATS.Client/Msg.cs
watfordgnf: I'm not sure sealing this class protects us in this case.

ColinSullivan1: Any suggestions?

src/Tests/UnitTests/TestMessageHeaders.cs
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        [Fact]
+        public void TestHeaderMultiValues()
+        {
+            var mh = new MsgHeader();
+            mh.Add("foo", "bar");
+            mh.Add("foo", "baz");
+
+            Assert.Equal("bar,baz", mh["foo"]);
```
kozlovic, line 160: I don't think this is correct. If there are 2 keys in the headers, you should return an array of strings, not a concatenated, comma separated, string.

src/NATS.Client/Msg.cs
kozlovic: I don't think this can be used to represent headers since multiple occurrence of the same key must be stored as an array of values. Also, you should have APIs because if the user tries to Get("foo-key"), you need internally to use the Canonical version: "Foo-Key".
Same when you are going to store the key, you need to store the Canonical version. For deserialization, you should ensure that the keys are canonical.
As previously discussed, the server for instance is not tolerant of a non canonical key. For JetStream and the de-dup feature, the server will look for a header called Msg-Id, and it if it is not stored this way, then that won't work.

kozlovic (changes requested):
I don't think you can get away with a simple map. You need to extend to make sure that keys are stored/retried in their canonical form.
Support for multiple values also means that key "Foo" may return an array of ["value1", "value2"] and not a single string of "value1, value2".
Finally, it is possible to receive a multiple line value. So this:
```
NATS/1.0\r\nTwo-Lines: this is\r\n a multilines value\r\n\r\n
```
needs to become "Tow-Lines": "this is a multilines value". Notice that the value spreads on 2 lines and start with a space:" a multilines..." the space will then be removed a concatenated to the previous line.

ColinSullivan1:
@kozlovic , great points - thanks!

Because MsgHeader is a subclass of `NameValueCollection`, one could use the [public virtual string[] GetValues (string name);](https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.namevaluecollection.getvalues?view=netcore-3.1) API to return the values as a string array.  The `NameValueCollection` class is a superclass for the .NET HTTP header classes - the string accessor returning comma separated values is an interesting feature that I used as a shortcut in testing.

I'll add tests for the `GetValues` APIs to be sure it works and will make sure multiline keys and values are handled correctly.

ColinSullivan1:
I can also add checks to make sure the keys are in canonical form.  The server handled the tests fine though...

kozlovic:
> I can also add checks to make sure the keys are in canonical form. The server handled the tests fine though...

Because in most cases it does not do anything with the headers. But there is currently 1 case where the server will read headers, it is for JS publish de-dup and in that case it will look for "Msg-Id" and so if the user sets to "msg-id" and library doesn't use canonical form, then server won't de-dup.

ripienaar:
yeah, server doesnt follow http spec, so gotta be careful. Also soon hopefully we'll add headers for tracing. I dont know .net but since you mentioned a hash map I'll also say that order of headers cant change - technically only the order of the same header more than once - but in general its probably a good idea not to tweak order

src/Tests/UnitTests/TestMessageHeaders.cs
ColinSullivan1: I've updated the example documentation and tests, look for `string[] values = mh.GetValues("foo");`

ColinSullivan1:
FYI, I've made various updates to cover what HTTP headers do, improve performance, and demonstrate multiple values in the documentation example and tests.

ColinSullivan1:
Are we g2g?

src/NATS.Client/Conn.cs
```diff
@@ -2003,7 +2009,7 @@ private int setMsgArgsAryOffsets(byte[] buffer, long length)
             int i = 0;
 
             // We only support 4 elements in this protocol version
```
watfordgnf: Is this still 4?

src/NATS.Client/Msg.cs
```diff
@@ -13,13 +13,355 @@
...
+                throw new ArgumentException("header", "header cannot be empty");
+            }
+            foreach (string s in header.Keys)
+            {
+                nvc[s] = header[s];
+            }
+        }
+
+        // buffer for parsing strings
+        public char[] stringBuf = new char[64];
```
watfordgnf: This should likely be internal.

watfordgnf (commented):
Saw one fix and had some minor comments.

src/NATS.Client/Msg.cs
ColinSullivan1: Oof - thanks - will fix!

watfordgnf (approved):
Looks good.

ColinSullivan1:
Thanks!  Good catch on the size of the args, appreciate it.  Turns out I didn't resize the array - embarrassing, but would have been much more so after a release.  :)

PR #402, opened 2020-07-28
Remove NoRandomize usage from example

jasper-d:
Remove NoRandomize usage from the clustered NATS example. My understanding is that randomization should be enabled in general (c.f. https://docs.nats.io/developing-with-nats/reconnect/random). Disabling it in the sample has caused some confusion with regards to the recommended settings.


watfordgnf (approved):

watfordgnf:
Good catch!

ColinSullivan1:
Thanks!

Issue #395, opened 2020-07-13
Mixing async and sync requests on the same connection doesn't work


tstenvi:
Using the same connection for sending async and sync requests does not work if the async request is sent prior to the synchronous one. The second request times out. The scenario works if the old request style option is set when creating the connection. The code below can be used to reproduce the problem. The issue has been reproduced using the 0.11.0-pre1 and 0.10.0 releases of the library.

```csharp
var cf = new ConnectionFactory();
var opts = ConnectionFactory.GetDefaultOptions();

var subject = Guid.NewGuid().ToString();

var subscribeCon = cf.CreateConnection();

void EventHandler(object sender, MsgHandlerEventArgs eventArgs)
{
    subscribeCon.Publish(eventArgs.Message.Reply, eventArgs.Message.Data);
}

var s = subscribeCon.SubscribeAsync(subject, EventHandler);

//opts.UseOldRequestStyle = true; // Uncomment this and it works
var requestConn = cf.CreateConnection(opts);

var m = await requestConn.RequestAsync(subject, Encoding.UTF8.GetBytes("Async request"), 10000);
Console.WriteLine(Encoding.UTF8.GetString(m.Data));

// This request time's out
m = requestConn.Request(subject, Encoding.UTF8.GetBytes("Sync request"), 10000);
Console.WriteLine(Encoding.UTF8.GetString(m.Data));
```

jasper-d:
Hi @tstenvi, I just pushed some changes here: https://github.com/nats-io/nats.net/pull/401
You may wanna give it a shot. It's not nice, it doesn't fix the underlying problem, but it __may__ avoid the deadlock I suspect here.

PR #404, opened 2020-07-30
Fix sync/async deadlock

jasper-d:
Fixes a deadlock #395 using TaskCreationOptions.RunContinuationsAsynchronously on netstandard1.6.
No integration test because I couldn't make it fail reliably (before the fix). 

Upgraded to net46 since #394 will do it anyways and it avoids an ugly workaround.

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -278,7 +278,7 @@ namespace NATS.Client
             {
                 if (!disposedValue)
                 {
-#if NET45
+#if NET46
                     if (executorTask != null)
                         executorTask.Dispose();
 #endif
@@ -421,7 +421,7 @@ namespace NATS.Client
 
             internal static void close(TcpClient c)
             {
-#if NET45
+#if NET46
                     c?.Close();
 #else
                     c?.Dispose();
@@ -636,7 +636,7 @@ namespace NATS.Client
                     // See Connection.deliverMsgs
                     Channel.close();
                     channelTask.Wait(500);
-#if NET45
+#if NET46
                     channelTask.Dispose();
 #endif
                     channelTask = null;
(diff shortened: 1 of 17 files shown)


watfordgnf:
I don't think its a dependency we can take, but [Microsoft's vs-threading repo](https://github.com/microsoft/vs-threading) has nice handlers for sync-over-async (aka JoinableTaskFactory) that perhaps would be a solution in this space.

jasper-d:
I don't know vs-threading, but looks like it wouldn't work for us indeed: https://github.com/microsoft/vs-threading/blob/master/src/Microsoft.VisualStudio.Threading/Microsoft.VisualStudio.Threading.csproj#L3, more specifically https://github.com/microsoft/vs-threading/search?q=RunContinuationsAsynchronously&unscoped_q=RunContinuationsAsynchronously (that's used by the JoinableTaskFactory).

ColinSullivan1 (approved):
LGTM.

ColinSullivan1:
Thanks!

src/NATS.Client/Conn.cs
```diff
@@ -2720,24 +2720,8 @@ private Msg requestSync(string subject, byte[] data, int offset, int count, int
...
-                    {
-                        // we *should* only have one, and it should be
-                        // a NATS timeout exception.
-                        throw e;
-                    }
-
-                    throw;
-                }
+                
+                return request.Waiter.Task.GetAwaiter().GetResult();
```
watfordgnf, line 2724: Does this still timeout after `timeout` milliseconds?

jasper-d: Yes, `timeout` is passed through `setupRequest` to `InflightRequest` where it's used to set the timeout.https://github.com/nats-io/nats.net/blob/ec0a317bde5cf0b7d88a5e63c83ae43527d0274c/src/NATS.Client/Internals/InFlightRequest.cs#L72

watfordgnf (approved):

PR #406, opened 2020-09-18
Support connecting to hostnames that resolve to IPv6 addresses

### LINKED ISSUE #386: Connect to IPv6 server with hostname provided
Hi I have trouble with connect to server via IPv6 if server address is provided as a hostname.

`var cf = new ConnectionFactory();`
`var c = cf.CreateConnection("nats://somehost.domain.net:4222");`

In this case GetInterNetworkAddressFamily method gets address family of address (it's AddressFamily.DNS) and returns null. Then default TcpClient is created (IPv4) and I can't connect to server.

Basically this issue is similar to  [https://github.com/danielwertheim/nats.net/commit/1c3bcba27e2be3d90e1c8147ffeed8506adc9beb](url), which is closed but is not fully fixed because it doesn't resolve host address from DNS. 
I think `Dns.GetHostAddresses(s.url.Host);.`  should be added to GetInterNetworkAddressFamily : 

```
var ipAddress = Dns.GetHostAddressesAsync(uri.Host).GetAwaiter().GetResult().FirstOrDefault();

            if (ipAddress == null)
            {
                throw new NATSConnectionException("Cannot resolve server endpoint");
            }

            return ipAddress.AddressFamily != AddressFamily.InterNetwork &&
                   ipAddress.AddressFamily != AddressFamily.InterNetworkV6
                ? (AddressFamily?)null
                : ipAddress.AddressFamily;
```
This issue reproduces only on .Net Core < 3.0 and on .Net Framework.

DavidSimner:
This PR fixes #386 and adds support for connecting to a hostname that resolves to an IPv6 address on .NET Framework (.NET Core already supported this).

The previous issue #272 that was addressed in #316 added support for connecting to an explicit IPv6 address.  This PR extends that work.


DavidSimner:
The build has failed, but the failing test: `IntegrationTests.TestBasic.TestAsyncInfoProtocolPrune` fails even before this PR, e.g. see https://github.com/nats-io/nats.net/runs/982062581 so I think that this PR is ready for review :)

ColinSullivan1:
Many thanks for the PR!  Much appreciated.  Yeah, that test flaps quite a bit; it's on the list to fix.  This resolves #386

ColinSullivan1 (approved):
LGTM.  Thanks!

PR #408, opened 2020-09-24
Fix typo in README.md

kruettlip:
Documentation contained "connection.Observere(subject)" but looking at the example below, it seems like "connection.Observe(subject)" would be correct.


ColinSullivan1 (approved):
LGTM - Thanks!

PR #412, opened 2020-10-06
Feature:  Lame Duck Mode Support

### LINKED ISSUE #399: Add Lame Duck Mode Support
Implement the .NET idiomatic equivalent of https://github.com/nats-io/nats.go/pull/572.

ColinSullivan1:
The NATS Server can enter lame duck mode, whereupon it sends an INFO message with "ldm" enabled to clients.  The NATS client library can notify applications, allowing them to clean up, drain their connection, and gracefully connect to another server.

This PR adds a connection event handler to support this functionality, along with associated tests.  While supported on every platform, tests are only performed on systems that support signaling (not windows).

Resolves #399

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -2336,17 +2336,10 @@ namespace NATS.Client
             info = ServerInfo.CreateFromJson(json);
             var discoveredUrls = info.connect_urls;
 
-            // The array could be empty/not present on initial connect,
-            // if advertise is disabled on that server, or servers that
+            // The discoveredUrls array could be empty/not present on initial
+            // connect if advertise is disabled on that server, or servers that
             // did not include themselves in the async INFO protocol.
-            // If empty, do not remove the implicit servers from the pool.
-            if (discoveredUrls?.Length == 0)
-            {
-                if (notify && info.ldm && opts.LameDuckModeEventHandler != null) {
-                    scheduleConnEvent(opts.LameDuckModeEventHandler);
-		        }
-                return;
-            }   
+            // If empty, do not remove the implicit servers from the pool.  
 
             // Note about pool randomization: when the pool was first created,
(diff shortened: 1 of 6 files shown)


src/NATS.Client/Conn.cs
```diff
@@ -2335,7 +2335,19 @@ internal void processInfo(string json, bool notifyOnServerAddition)
...
+            // if advertise is disabled on that server, or servers that
+            // did not include themselves in the async INFO protocol.
+            // If empty, do not remove the implicit servers from the pool.
+            if (discoveredUrls?.Length == 0)
+            {
+                if (notify && info.ldm && opts.LameDuckModeEventHandler != null) {
+                    scheduleConnEvent(opts.LameDuckModeEventHandler);
+		        }
+                return;
+            }
```
matthiashanel: if the next block on line 2355 gets skipped because lenght == 0 then line 2368 will be executed.
Together this does the same as this entire block, which makes it redundant. Delete?

matthiashanel (changes requested):
one condition appears duplicated

src/NATS.Client/Conn.cs
ColinSullivan1: It's in case discovered urls are empty.. so if the client is ldm'd and receives empty urls it doesn't lose what it has...

matthiashanel: but the next check here is ` if (discoveredUrls != null && discoveredUrls.Length > 0)` which is identical.
if discoveredUrls is empty we wouldn't take that if and skip the block.
Then we would execute line 2638 `if (notify && info.ldm && opts.LameDuckModeEventHandler != null)`
This is identical to the code inside this if.
next the function is over, taking the place of return.

matthiashanel (approved):
LGTM

ColinSullivan1:
Thanks for the review @matthiashanel.

PR #413, opened 2020-10-09
Add No-Responders Feature

### LINKED ISSUE #400: Add support for no responders error on requests
Implement the equivalent of https://github.com/nats-io/nats.go/pull/576.

If connected to a new server and a request is sent when there are no responders the request will return as soon as a response from the server is received. The response will have an empty payload and a 503 status.

Depends on #394.

ColinSullivan1:
## Add the No Responders Client Feature

Resolves #400 

When a request is made from the client, if the server detects there is no interest in the request subject (no subscribers), it will immediately return a status header indicating there are no responders.  When this happens, the client will throw a new `NATSNoRespondersException` indicating there are no responders available for a request.  This circuit breaking behavior allows applications to avoid waiting for timeouts for requests that NATS can detect will fail.

The `NATSNoRespondersException` is a subclass of the `NATSTimeoutException` in an effort to maintain some backward compatibility.

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -2757,9 +2757,7 @@ namespace NATS.Client
             InFlightRequest request;
             bool isClosed;
 
-            var m = args.Message;
-
-            if (m == null)
+            if (args.Message == null)
                 return;
 
             var subject = args.Message.Subject;
@@ -2800,13 +2798,13 @@ namespace NATS.Client
 
             if (!isClosed)
             {
-                if (IsNoRespondersMsg(m))
+                if (IsNoRespondersMsg(args.Message))
                 {
                     request.Waiter.TrySetException(new NATSNoRespondersException());
                 }
                 else
                 {
-                    request.Waiter.TrySetResult(m);
+                    request.Waiter.TrySetResult(args.Message);
                 }
             }
             else
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -2747,7 +2747,9 @@ namespace NATS.Client
 
         private static bool IsNoRespondersMsg(Msg m)
         {
-            return m != null && m.HasHeaders && MsgHeader.noResponders.Equals(m.Header[MsgHeader.Status]);
+            return m != null && m.HasHeaders &&
+                MsgHeader.noResponders.Equals(m.Header[MsgHeader.Status]) &&
+                m.Data.Length == 0;
         }
 
         private void RequestResponseHandler(object sender, MsgHandlerEventArgs args)
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -2745,15 +2745,22 @@ namespace NATS.Client
             }
         }
 
-        private void RequestResponseHandler(object sender, MsgHandlerEventArgs e)
+        private static bool IsNoRespondersMsg(Msg m)
+        {
+            return m != null && m.HasHeaders && MsgHeader.noResponders.Equals(m.Header[MsgHeader.Status]);
+        }
+
+        private void RequestResponseHandler(object sender, MsgHandlerEventArgs args)
         {
             InFlightRequest request;
             bool isClosed;
 
-            if (e.Message == null)
+            var m = args.Message;
+
+            if (m == null)
                 return;
 
-            var subject = e.Message.Subject;
+            var subject = args.Message.Subject;
 
             lock (mu)
             {
@@ -2791,7 +2798,14 @@ namespace NATS.Client
 
             if (!isClosed)
             {
-                request.Waiter.TrySetResult(e.Message);
+                if (IsNoRespondersMsg(m))
+                {
+                    request.Waiter.TrySetException(new NATSNoRespondersException());
+                }
+                else
+                {
+                    request.Waiter.TrySetResult(m);
+                }
             }
             else
             {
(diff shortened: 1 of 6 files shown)


src/Tests/IntegrationTests/TestBasic.cs
```diff
@@ -614,6 +618,37 @@ public void TestRequestWithOffset()
...
+                Options opts = Context.GetTestOptions();
+
+                using (var c = Context.ConnectionFactory.CreateConnection(opts))
+                {
+                    var sw = Stopwatch.StartNew();
+                    Assert.Throws<NATSNoRespondersException>(() => c.Request("foo", null, 10000));
+                    sw.Stop();
+
+                    // make sure we didn't also time out.
+                    Assert.True(sw.ElapsedMilliseconds < 5000);
```
matthiashanel, line 635: wouldn't you see this as a problem?
Before you'd sit there, thus essentially sleeping.
Now your existing code starts to spin

matthiashanel (commented):
still working my way through, but got one comment

src/Tests/IntegrationTests/TestBasic.cs
ColinSullivan1: Not sure I understand...  I can see the comment being misleading through.  The request should circuit break by the server's response w/ error code and return much earlier than the 10s configured...  The check is < 5 seconds generous for really, really flaky CI, it's really just checking that it returned earlier than the configured timeout.

ColinSullivan1: Ahh - are you saying because NATSNoRespondersException is a subclass of the timeout exception, existing application code could spin if someone had requests in a tight loop...

I wrestled with that a bit.  Throw a new exception (breaking existing code), vs subclassing allowing existing code to succeed but possibly churn.   Tough call.

matthiashanel: I'd say sending a request in a loop is currently slowed down because on timeout we will have waited.
No need for any extra application logic to do that then.
This does not appear to be opt in... which now would start to cause applications to not wait.

src/NATS.Client/Conn.cs
```diff
@@ -2745,15 +2745,24 @@ private void RemoveOutstandingRequest(string requestId)
...
             bool isClosed;
 
-            if (e.Message == null)
+            var m = args.Message;
+
+            if (m == null)
                 return;
 
-            var subject = e.Message.Subject;
+            var subject = args.Message.Subject;
```
matthiashanel: args.Message is also m. use m throughout or delete m

src/NATS.Client/Exceptions.cs
```diff
@@ -179,4 +179,16 @@ public class NATSConnectionDrainingException : NATSConnectionException
     {
         public NATSConnectionDrainingException() : base("Connection is draining.") { }
     }
+
+    /// <summary>
+    /// The exception thrown when the server has detected there are no responders for a request.
+    /// /// </summary>
```
matthiashanel: extra ///

src/Tests/IntegrationTests/TestBasic.cs
ColinSullivan1: Similar problem in go... ;).  However, the difference is nuanced - in go it's a separate error, here it's a subclass.  Are you suggesting clients opt-in?

src/NATS.Client/Msg.cs
```diff
@@ -126,25 +134,59 @@ internal MsgHeader(byte[] bytes, int byteCount) : base()
...
+            int kvStart = HeaderLen;
+
+            // We need to be fast so do a byte comparison.
             int i;
             for (i = 0; i < HeaderLen; i++)
             {
                 if (bytes[i] != HeaderBytes[i])
                 {
-                    throw new NATSInvalidHeaderException();
+                    // check for inlined status:  e.g. "NATS/1.0 503\r\n\r\n"
```
matthiashanel, line 150: should we drop the header if 1.0 becomes smth this version of the library does not understand?

matthiashanel (changes requested):
Minor changes

src/Tests/IntegrationTests/TestBasic.cs
matthiashanel: I'm not saying that the problem does not exist in go either. It's a problem for all code already written.

src/NATS.Client/Conn.cs
ColinSullivan1: Will fix...  Thanks.

src/NATS.Client/Msg.cs
ColinSullivan1: Right now the client will reject the message having an invalid header.

ColinSullivan1: I think that behavior is OK because a change in header version would indicate a breaking change - which should be addressed by updating the client version.  wdyt?

matthiashanel (commented):
Code looks good

src/NATS.Client/Msg.cs
matthiashanel: this is not part of this change so probably can be addressed in a separate pr.

Yes it would, but do you really want to risk exposing running clients (not updated yet) to such a header?

matthiashanel: If we plan for changes once we introduce them we haven't planed for them at all.

matthiashanel (approved):
LGTM.

header version and disabling this feature can be separate PR

PR #421, opened 2020-11-08
simple `async` poc

AndyPook:
this is just for discussion..
a demonstration of using System.Threadding.Channels to make Nats.Client.Channel allow an async mode. This should a subscriber to not need to block a Thread and hopefully allow the case where having many subscribers in one process does not flood the ThreadPool.

I have tried to keep this "minimal". It only really affects the SyncSubscriber. Further/deeper refactoring could simplify the Async/Sync versions into one. It should be possible to also maintain the existing api to retain backward compat.
There is a lot of other refactoring that could follow to


src/NATS.Client/Channel.cs
```diff
@@ -86,31 +88,58 @@ internal void reset()
         }
     }
 
+    public class Channel<T> : AsyncChannel<T> { }
```
AndyPook: done this way just to make it simple to switch between `SyncChannel` (existing code) and the `AsyncChannel` (proposal)

src/NATS.Client/Conn.cs
```diff
@@ -3917,15 +3917,7 @@ internal void removeSubSafe(Subscription s)
...
-            if (s.mch != null)
-            {
-                if (s.ownsChannel)
-                    s.mch.close();
-
-                s.mch = null;
-            }
-
-            s.closed = true;
+            s.CloseSafe();
```
AndyPook: a simple example of making class internals "safe" rather than exposing everything as `internal`

src/NATS.Client/NATS.Client.csproj
```diff
@@ -39,5 +39,6 @@
 
   <ItemGroup>
     <PackageReference Include="System.Collections.Specialized" Version="4.3.0" />
+    <PackageReference Include="System.Threading.Channels" Version="4.7.1" />
```
AndyPook, line 42: STC is ns1.3 so should be fine for net46 etc

src/NATS.Client/Subscription.cs
```diff
@@ -41,7 +41,7 @@ public class Subscription : ISubscription, IDisposable
         internal bool closed = false;
         internal bool connClosed = false;
 
-        internal Channel<Msg> mch = null;
+        protected Channel<Msg> mch = null;
```
AndyPook: related to `CloseSafe()`

src/NATS.Client/SyncSub.cs
```diff
@@ -128,5 +130,52 @@ public Msg NextMessage(int timeout)
 
             return msg;
         }
+
+        public async ValueTask<Msg> NextMessageAsync(CancellationToken cancellationToken = default)
```
AndyPook, line 128: `async` version of `NextMessage`.
I believe that the "snapshot" of the locals and locking is not actually necessary as I would only expect NextMessage to be called sequentially within a loop (?). But I am sure the devil is in the details.

src/Tests/IntegrationTests/TestBasic.cs
```diff
@@ -268,6 +268,34 @@ public void TestSyncSubscribe()
             }
         }
 
+        [Fact]
+        public async Task TestSyncSubscribeAsync()
```
AndyPook, line 272: simple async version of `TestSyncSubscribe"

AndyPook:
I have noted that some of the "old style request" tests are failing.
Some of these are because it now sometimes runs too fast (by a few ms). This is intermittent. Sometimes it will pass.
If we go ahead I suspect that some of this will come out in the wash as we get deeper into it.

jasper-d:
The tests are notorious for being flaky, I wouldn't worry about them.

AndyPook:
I suspected as much. Thanks for confirming though.

Do you have any thoughts on what the appetite might be for a more intrusive refactor would be? How close does this need to be to the go client? I can see how keeping it close would make keeping this up-to-date somewhat easier 🤔

src/NATS.Client/SyncSub.cs
sixlettervariables: Is ValueTask available in 4.6?

AndyPook: I think it does because System.Threading.Channels (which needs it) targets netstandard1.3 so should work fine with net46.

(I have a note to investigate, just haven't dug in deep enough yet)

sixlettervariables: MSDN says .NET Standard 2.1: https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask-1

jasper-d: It comes with System.Threading.Task.Extensions through nuget. The PR was build targeting net46 and netstandard1.6 just fine: https://dev.azure.com/NATS-CI/NATS-CI/_build/results?buildId=426&view=logs&j=f218f9e9-0264-5a0f-6bf7-3564b2c4a2c7&t=db6d43a5-26f0-51c1-2344-796106ef1855

jasper-d:
I trace the following code to see how it behaves (targeting netcoreapp3.1): https://github.com/jasper-d/nats.net/commit/f7eb37097eec19e2dc23887be7017368c3db8803#diff-a2ea9c69488db905f5a7364186cf4824979127b8cb23f86cf92704674acb31b9

Using `ISyncSubscription.NextMessageAsync` the trace looks just fine. Perfview reports 21 `Microsoft-Windows-DotNETRuntime/AppDomainResourceManagement/ThreadCreated` events (for 1000 subscriptions). CPU usage is virtually non-existing, I didn't measure it though. 

For `IAsyncSubscription.SubscribeAsync` Perfview reports 1019 `Microsoft-Windows-DotNETRuntime/AppDomainResourceManagement/ThreadCreated` events, so 1 for every subscription plus some for good measure. I don't know if thats a regression, haven't traced it using upstream yet. CPU load is quite high (like 10%-50% on my old 4C/8T notebook). I suspect thats because there are 1000+ threads being sampled (see screenshot) + plus some overhead of those threads spin-waiting + thread switching.


image

jasper-d:
> For IAsyncSubscription.SubscribeAsync Perfview reports 1019 Microsoft-Windows-DotNETRuntime/AppDomainResourceManagement/ThreadCreated events, so 1 for every subscription plus some for good measure. I don't know if thats a regression, haven't traced it using upstream yet.

Not a regression, current implementation shows 1017 threads.

> Do you have any thoughts on what the appetite might be for a more intrusive refactor would be? How close does this need to be to the go client? I can see how keeping it close would make keeping this up-to-date somewhat easier 🤔

I would like to know that too. I have some ideas how nats.net could be improved (and wouldn't mind contributing to it) but keeping things compatible with .NET FX and aligned  with nats.go would definitely a challenge. On the other hand diverging from it could possible encourage contributions, I guess. Do you have any thoughts about that @ColinSullivan1 @watfordgnf?

src/NATS.Client/Channel.cs
```diff
@@ -335,7 +364,77 @@ internal int Count
...
+        readonly ChannelWriter<T> writer;
+        readonly ChannelReader<T> reader;
+
+        internal AsyncChannel()
+        {
+            Name = "Unnamed channel " + this.GetHashCode();
+            var channel = Channel.CreateUnbounded<T>(new UnboundedChannelOptions
+            {
+                SingleWriter = true,
+                SingleReader = true
```
jasper-d: Wouldn't `SingleReader = true` be an issue if clients call `NextMessageAsync` concurrently on the same subscription?

AndyPook: true.
this was just a cop out  to get to the point where _something_ would work. This is just to stimulate some conversation :)

There is some thoughts that the behavior of the two current Subscriber types is different. Async is by definition SingleReader. The Sync it is possible to have multiple readers (just not the style I've been using.

I have a branch that actually gets rid of N.C.Channel altogether. Also hugely simplifies the Subscription types and the interaction with Connection. With this refactoring the Channel type is no longer needed. Just folds the S.T.Channel into a single impl of Subscription.

I haven't had much time recently. Once I have something that "works" I'll post another PR for discussion.

AndyPook:
@jasper-d  please note that I have not put any focus into how the Async Subscription behaves (other than to keep it working).

As mentioned in the Channel thread, the way Channel, Subscription & Connection are currently factored makes it somewhat tricky to keep everything "lined up".
The Channel "ownership" makes thing awkward. For Async the Connection runs the subscription message consumer (via some double dispatch). Whereas the Sync version creates the Channel which subscribers pull from.

The theory I am looking at separately is to have a single Subscription type. The Connection just calls `add` regardless of what type of sub it is. The Async version runs it's own consumer reading from the private S.T.Channel, dispatching to the EventHandler. But also has NextMessage/NextMessageAsync for the Sync/pull type consumer.
I think this keeps the separation of concerns a bit neater. And also allows for a lot of refactor-by-deletion (which always makes me happy). Another nice side-effect is that is allows the removal of a lot of the locking/Monitoring and need for `internal` as the ownership and updating of members is a lot easier to reason about.

The hope is that these bits provide a nice boundary between the network/protocol bits in Connection and the api most users interact with.

At least that's the theory.
If I can put out some fires at work, I may have something to show fairly soon.

AndyPook:
@jasper-d  I meant to add...
I'll try to add the equivalent you your TestSubs into a Benchmark.net suite. It'd be super interesting (to me) to capture allocations, threads etc to have some actual empirical data (geek!)

jasper-d:
> @jasper-d please note that I have not put any focus into how the Async Subscription behaves (other than to keep it working).

I know, I just wanted to point out what huge difference your changes make in terms of efficiency and double check that the "Async" Subs aren't negatively affected (I was a bit concerned about the spin wait). :) 

> As mentioned in the Channel thread, the way Channel, Subscription & Connection are currently factored makes it somewhat tricky to keep everything "lined up".
> The Channel "ownership" makes thing awkward. For Async the Connection runs the subscription message consumer (via some double dispatch). Whereas the Sync version creates the Channel which subscribers pull from.
> 
> The theory I am looking at separately is to have a single Subscription type. The Connection just calls `add` regardless of what type of sub it is. The Async version runs it's own consumer reading from the private S.T.Channel, dispatching to the EventHandler. But also has NextMessage/NextMessageAsync for the Sync/pull type consumer.
> I think this keeps the separation of concerns a bit neater. And also allows for a lot of refactor-by-deletion (which always makes me happy). Another nice side-effect is that is allows the removal of a lot of the locking/Monitoring and need for `internal` as the ownership and updating of members is a lot easier to reason about.
> 
> The hope is that these bits provide a nice boundary between the network/protocol bits in Connection and the api most users interact with.

To summarize, you plan to use S.T.Channel to bridge sync/async (as you did here already) as well as using it as a unified interface for both "Async" and Sync subscriptions and Connection on the other end. Did I get that right?
I looked multiple times at ways to add async capabilities, but shied away being afraid to break stuff in Conn.cs, concluding that the only feasible way would a seperate async implementation of IO dependent paths. So kudos to you.

>I'll try to add the equivalent you your TestSubs into a Benchmark.net suite. It'd be super interesting (to me) to capture allocations, threads etc to have some actual empirical data (geek!)

That would be interesting, indeed. Currently benchmarking using BDN is not really feasible, because as you said, there is a lot of stuff intertwined and afaik currently you simply cannot benchmark anything without taling to nats-server (which will skew the results for multiple reasons). So if your refactoring opens up a possibility for more granular benchmarks that would be another benifit.

AndyPook:
@jasper-d  having another taking a look and verifying from their pov is very much appreciated :)
those call stats are interesting.

Yes, that description is pretty much correct.
It will be tricky to completely verify 100% equivalent behavior. But I think that having an easier to see boundary between the network stream parsing bits and the higher level Subscription parts will make it easier to verify that the new behavior is "correct" (maybe, probably... we'll see 😬  )
The difficult thing is in doing sync-over-async. Things work "better" if it's async all the way through. But that's a bit of a challenge for some projects. The SpinWait thing I'm using a not "ideal", again just a technique to get something to work. I would expect a different impl if this goes forward.

My theory is that the Subscription layer is where we want the experience to be as close to dotnet idiomatic as possible. The underlying network shenanigans can remain closer to the go stuff (if that's what is desired).

What I have in my head if that the boundary is just the `add` method (for subs). And keeping the existing "Sync"/"Async" api's but making it possible for others with `async` overloads etc

AndyPook:
took a bit of a leap 😬 
- removed Channel altogether
- refactored Subscription to include S.T.Channel
- AsyncSub now "owns" the consumer task
- Conn just `addMessage` to Subs. no diff between types. no consumer loop. no SubChannelPool

lots going on here.
Most test pass
The "old request Async cancellation" tests all seem to hang, looking into it.
"close"ing & restarting Subs needs proving,

Still a long way to go

AndyPook:
I have solutions for the "old request" problems. Lots more refactor-by-deletion which feels good :)
should be able to push in the next few days
(day job getting in the way)

AndyPook:
fixed a few things. All test should pass now 🤞 

- simplified oldRequest
- waits for Async sub to start (avoids Start/Publish race)
- check for "closed" in GetMessage and throw

I'm still a little confused about the semantics of `conn = null` in AsyncSub. Not sure why it would ever do this 🤔 

Anyway, I think this is a good point to get some feedback about this approach ?

src/NATS.Client/Conn.cs
```diff
@@ -3131,82 +2962,51 @@ public Msg Request(Msg message)
...
-            return Task.Run(() =>
+            // check if we are already cancelled.
+            ct.ThrowIfCancellationRequested();
+
+            var originalCt = ct;
+            if (timeout > 0)
             {
-                // check if we are already cancelled.
-                ct.ThrowIfCancellationRequested();
+                var cts = CancellationTokenSource.CreateLinkedTokenSource(ct);
```
watfordgnf: `cts` will need to be disposed.

AndyPook: thanks for taking a look
hopefully b7a55b5 resolves that?

scottf:
Should be address by V2

PR #422, opened 2020-11-08
comma issue

scottf:
So I built some test code to understand how the .Net code writes the headers, and I think there is a problem if the value contains a comma. Consider this header:
```
Header1: Foo
```
When this is serialized, the result will look like:
```
Header1: Foo␍␊
```

Now consider this data where the key has multiple values.
```
Header1: Foo
Header1: Comma,Inside
```

When this is serialized the result is:
```
Header1: Foo,Comma,Inside␍␊
```
My guess is that when this de-serializes the header ends up as:
```
Header1: Foo
Header1: Comma
Header1: Inside
```

which is not what is desired. This could be circumvented by changing the serialization as in the proposed code which would serialize like this 
```
Header1: Foo␍␊Header1: Comma,Inside␍␊
```

### DIFF
--- a/src/NATS.Client/Msg.cs
+++ b/src/NATS.Client/Msg.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2015-2020 The NATS Authors
+// Copyright 2015-2020 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


scottf:
@ColinSullivan1 So I found something out while testing. Now that I can publish I sent a request with a header with 2 values `one` and `two,three` . The go client received this as 3 headers, `one`, `two` and `three`

ColinSullivan1:
OK, I'll check this out, thanks Scott.

scottf:
Single line is fine. There are other issues with headers that need to be addressed. New issue for this: https://github.com/nats-io/nats.net/issues/448

scottf:
Fix not necessary, will be addressed in https://github.com/nats-io/nats.net/issues/448

PR #423, opened 2020-11-10
Fix client version

### LINKED ISSUE #415: Fix client version in protocol.
Client indicates it is 0.0.1 to the NATS server, this needs to be fixed.

jasper-d:
Fixes #415

### DIFF
--- a/src/NATS.Client/NATS.cs
+++ b/src/NATS.Client/NATS.cs
@@ -71,7 +71,7 @@ namespace NATS.Client
         /// <summary>
         /// Client version
         /// </summary>
-        public static string Version = typeof(Defaults).GetTypeInfo().Assembly.GetName().Version.ToString();
+        public static readonly string Version = typeof(Defaults).GetTypeInfo().Assembly.GetName().Version.ToString();
 
         /// <summary>
         /// The default NATS connect url ("nats://localhost:4222")
--- a/src/NATS.Client/NATS.cs
+++ b/src/NATS.Client/NATS.cs
@@ -12,6 +12,7 @@
 // limitations under the License.
 
 using System;
+using System.Reflection;
 
 /*! \mainpage %NATS .NET client.
  *
@@ -70,7 +71,7 @@ namespace NATS.Client
         /// <summary>
         /// Client version
         /// </summary>
-        public const string Version = "0.0.1";
+        public static string Version = typeof(Defaults).GetTypeInfo().Assembly.GetName().Version.ToString();
 
         /// <summary>
         /// The default NATS connect url ("nats://localhost:4222")


src/NATS.Client/NATS.cs
```diff
@@ -70,7 +71,7 @@ public static class Defaults
         /// <summary>
         /// Client version
         /// </summary>
-        public const string Version = "0.0.1";
+        public static readonly string Version = typeof(Defaults).GetTypeInfo().Assembly.GetName().Version.ToString();
```
watfordgnf, line 74: Does this work if you pack the assembly into a single executable?

jasper-d: Good question, [docs ](https://docs.microsoft.com/en-us/dotnet/core/deploying/single-file#api-incompatibility) kinda indicate it does but I will test that using a single file deployment targeting net5.0 to be sure.

watfordgnf: I had issues doing that with a project at work (among other problems with packed executables).

jasper-d: Out of curiosity, did you have issues with assembly version specifically or was it about reflection in general/assembly location information? 

Tested it using the following steps:

1. Built nuget package of NATS.Client, setting version to 3.2.1 and published it to a local feed
2. Created a new console application with the following .csproj and Program.cs
```
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net5.0</TargetFramework>
    <RootNamespace>test_app</RootNamespace>
    <Version>42.42.42</Version>
    <PublishSingleFile>true</PublishSingleFile>
    <SelfContained>true</SelfContained>
    <RuntimeIdentifier>linux-x64</RuntimeIdentifier>
    <PublishTrimmed>true</PublishTrimmed>
    <PublishReadyToRun>true</PublishReadyToRun>
    <PublishReadyToRunShowWarnings>true</PublishReadyToRunShowWarnings>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="nats.client" Version="3.2.1" />
  </ItemGroup>
```
```
static class Program {static void Main() => System.Console.WriteLine($"NATS.Client Version: {NATS.Client.Defaults.Version}");}
```
3. `dotnet publish -c Release --force -o ./single-file-app`
4. `./single-file-app/test-app`
Output is `NATS.Client Version: 3.2.1.0`

The R2R compiler produces most of the times (but not always) many warnings, but that appears to be more of a general problem. To me the entire single file/r2r story looks still rather "fresh".

watfordgnf: I think conceptually the code you have should work, but as you stated single file is still "fresh".

watfordgnf (approved):

ColinSullivan1 (approved):
LGTM.  Thanks!

PR #432, opened 2021-02-15
Fix TCP connectivity on systems w/o IPv6 support

### LINKED ISSUE #431: TCP connectivity broken on some systems
Release 0.11.0 introduced a change that broke TCP connections on systems without IPv6/dual stack support.
This does not affect systems where the IPv6 stack is present but address assignment is disabled, but systems where IPv6 isn't available at all (i.e. disabled through Kernel parameter `ipv6_disable=1` at boot time).
On such systems connections to nats-server will fail with a `NATSNoServersException` without much more information.

Possible related but most likely not fixed by my PR: #425

jasper-d:
- Fix TCP connectivity
- Throw specific exceptions in case of a connection failure
  - please review carefully, there's some global state involved that makes me feel uncomfortable

Fixes #431


src/NATS.Client/Conn.cs
```diff
@@ -377,8 +377,10 @@ internal void open(Srv s, int timeoutMillis)
                         sslStream = null;
                     }
 
-                    client = new TcpClient(AddressFamily.InterNetworkV6);
-                    client.Client.DualMode = true;
+                    client = new TcpClient(Socket.OSSupportsIPv6 ? AddressFamily.InterNetworkV6 : AddressFamily.InterNetwork);
+                    if(Socket.OSSupportsIPv6)
```
watfordgnf:
```suggestion
                    if (Socket.OSSupportsIPv6)
```

src/NATS.Client/Conn.cs
```diff
@@ -811,8 +813,9 @@ private void setupServerPool()
         // createConn will connect to the server and wrap the appropriate
         // bufio structures. It will do the right thing when an existing
         // connection is in place.
-        private bool createConn(Srv s)
+        private bool createConn(Srv s, out Exception ex)
```
watfordgnf, line 816: Should probably use ExceptionDispatchInfo so you can preserve the callstack below.

watfordgnf: Alternatively, ExceptionDispatchInfo.Throw(exToThrow) down below but that excludes .NET Framework.

jasper-d: > Alternatively, ExceptionDispatchInfo.Throw(exToThrow) down below but that excludes .NET Framework.

Unfortunately that's neither available in netcoreapp < 2.0 nor netstandard < 2.1.

> Should probably use ExceptionDispatchInfo so you can preserve the callstack below.

Been there, reverted it. There are two problems with `ExceptionDispatchInfo` here:
1. In the `doReconnect` path `Connection.lastEx` is set which is a raw exception. One could unwrap `ExceptionDispatchInfo.SourceExcpetion` in the reconnect code, but would probably loose the stack trace in the process. Changing `Connection.lastEx` to `ExceptionDispatchInfo` is not feasible, since it would result in a refactoring of large parts of Conn.cs and one would still hit some dead ends (i.e. `ConnEventArgs` exposes some exceptions though its `Error` property, so unwrapping there would be mandatory).
2. Even when unwrapping in the `doReconnect` path, we couldn't just `edi.Throw()` because that would add a bunch of new exceptions which clients are not expected to handle right now. So we could either do something like:
```
if(exToThrow?.SourceException is NATSException) exToThrow.Throw();
if(exToThrow is object) throw new NATSConnectionException("Failed to connect", exToThrow.SourceException) // <-- some stack trace lost
throw new NATSNoServersException("");
```
or even worse
```
if(exToThrow?.SourceException is NATSException) exToThrow.Throw();
if(exToThrow is object) 
  try { exToThrow.Throw(); }
  catch (Exception ex) { throw new NATSConnectionException("Failed to connect", ex); } // <-- stack trace preserved but a kitten died
throw new NATSNoServersException("");
```


While it's not perfect, I think this implementation is an acceptable compromise. It provides better information regarding the actual error and keeps changes limited.
For reference stack traces of socket and timeout exceptions using the implementation in this PR and w/ EDI + unwrapping `SourceExcpetion` (may look different on legacy .NET):

This PR:
```
NATS.Client.NATSConnectionException: Failed to connect
---> System.Net.Sockets.SocketException (97): Address family not supported by protocol
   at System.Net.Sockets.Socket..ctor(AddressFamily addressFamily, SocketType socketType, ProtocolType protocolType)
   at System.Net.Sockets.TcpClient.InitializeClientSocket()
   at System.Net.Sockets.TcpClient..ctor(AddressFamily family)
   at NATS.Client.Connection.TCPConnection.open(Srv s, Int32 timeoutMillis) in /src/nats.net/src/NATS.Client/Conn.cs:line 380
   at NATS.Client.Connection.createConn(Srv s, Exception& ex) in /src/nats.net/src/NATS.Client/Conn.cs:line 823
   --- End of inner exception stack trace ---
   at NATS.Client.Connection.connect() in /src/nats.net/src/NATS.Client/Conn.cs:line 1186
   at NATS.Client.ConnectionFactory.CreateConnection(Options opts) in /src/nats.net/src/NATS.Client/ConnectionFactory.cs:line 171
   at IpTests.Program.Main(String[] args) in /src/nats.net/src/Samples/IpTests/Program.cs:line 27
   
NATS.Client.NATSConnectionException: timeout
   at NATS.Client.Connection.connect() in /src/nats.net/src/NATS.Client/Conn.cs:line 1185
   at NATS.Client.ConnectionFactory.CreateConnection(Options opts) in /src/nats.net/src/NATS.Client/ConnectionFactory.cs:line 171
   at IpTests.Program.Main(String[] args) in /src/nats.net/src/Samples/IpTests/Program.cs:line 27

```
With EDI + unwrapping:
```
NATS.Client.NATSConnectionException: Failed to connect
---> System.Net.Sockets.SocketException (97): Address family not supported by protocol
   at System.Net.Sockets.Socket..ctor(AddressFamily addressFamily, SocketType socketType, ProtocolType protocolType)
   at System.Net.Sockets.TcpClient.InitializeClientSocket()
   at System.Net.Sockets.TcpClient..ctor(AddressFamily family)
   at NATS.Client.Connection.TCPConnection.open(Srv s, Int32 timeoutMillis) in /src/nats.net/src/NATS.Client/Conn.cs:line 381
   at NATS.Client.Connection.createConn(Srv s, ExceptionDispatchInfo& ex) in /src/nats.net/src/NATS.Client/Conn.cs:line 824
   --- End of inner exception stack trace ---
   at NATS.Client.Connection.connect() in /src/nats.net/src/NATS.Client/Conn.cs:line 1187
   at NATS.Client.ConnectionFactory.CreateConnection(Options opts) in /src/nats.net/src/NATS.Client/ConnectionFactory.cs:line 171
   at IpTests.Program.Main(String[] args) in /src/nats.net/src/Samples/IpTests/Program.cs:line 27

NATS.Client.NATSConnectionException: timeout
   at NATS.Client.Connection.TCPConnection.open(Srv s, Int32 timeoutMillis) in /src/nats.net/src/NATS.Client/Conn.cs:line 395
   at NATS.Client.Connection.createConn(Srv s, ExceptionDispatchInfo& ex) in /src/nats.net/src/NATS.Client/Conn.cs:line 824
--- End of stack trace from previous location where exception was thrown ---
   at NATS.Client.Connection.connect() in /src/nats.net/src/NATS.Client/Conn.cs:line 1185
   at NATS.Client.ConnectionFactory.CreateConnection(Options opts) in /src/nats.net/src/NATS.Client/ConnectionFactory.cs:line 171
   at IpTests.Program.Main(String[] args) in /src/nats.net/src/Samples/IpTests/Program.cs:line 27
```

ColinSullivan1 (approved):
This looks good to me.  @watfordgnf  - look OK?

watfordgnf (approved):

PR #434, opened 2021-02-20
[Added] CreateUserSeed and CreateAccountSeed so these can be generated

matthiashanel:
Signed-off-by: Matthias Hanel <mh@synadia.com>

altered visibility for some functions for testing.
Since nkeys don't store the type I am not returning an NKEY and return the seed instead.
That seed can then be used to create the nkey.

### DIFF
--- a/src/NATS.Client/NKeys.cs
+++ b/src/NATS.Client/NKeys.cs
@@ -12,6 +12,8 @@
 // limitations under the License.
 using NATS.Client.NaCl;
 using System;
+using System.IO;
+using System.Text;
 
 namespace NATS.Client
 {
@@ -231,8 +233,7 @@ namespace NATS.Client
                 src.Remove(0);
         }
 
-
-        private static byte[] DecodeSeed(byte[] raw)
+        internal static byte[] DecodeSeed(byte[] raw)
         {
             // Need to do the reverse here to get back to internal representation.
             byte b1 = (byte)(raw[0] & 248);  // 248 = 11111000
@@ -261,7 +262,7 @@ namespace NATS.Client
             }
         }
 
-        private static byte[] DecodeSeed(string src)
+        internal static byte[] DecodeSeed(string src)
         {
             return DecodeSeed(Nkeys.Decode(src));
         }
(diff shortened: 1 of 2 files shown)


src/NATS.Client/NKeys.cs
```diff
@@ -284,5 +285,59 @@ static public NkeyPair FromSeed(string seed)
...
+        public static string CreateUserSeed()
+        {
+            return CreateSeed(PrefixByteUser);
+        }
+
+        /// <summary>
+        /// Creates a private account seed String.
+        /// </summary>
+        /// <returns>A NATS Ed25519 Account Seed</returns>
+        public static string CreateAccountSeed()
```
ColinSullivan1, line 338: Perhaps add the others for completeness in case someone want to write a utility out of this... wdyt?

eg. 
```csharp
public static string CreateOperatorSeed();
public static string CreateServerSeed();
public static string CreateClusterSeed();
```

src/Tests/UnitTests/TestNkeys.cs
```diff
@@ -0,0 +1,56 @@
...
+using NATS.Client;
+using Xunit;
+
+namespace UnitTests
+{
+#pragma warning disable CS0618
+    public class TestNkeys
+    { 
+        [Fact]
+        public void TestNKEYEncodeDecode()
```
ColinSullivan1, line 24: Maybe add seeds created by CreateUserSeed, CreateAccountSeed, etc...

ColinSullivan1 (commented):
LGTM so far.  Thanks @matthiashanel!  A few requests, certainly open for discussion.

src/NATS.Client/NKeys.cs
matthiashanel: will add CreateOperatorSeed. Server and Cluster are not used.

src/Tests/UnitTests/TestNkeys.cs
matthiashanel: already have TestNKEYCreateUserSeed and TestNKEYCreateAccountSeed further down, will add the operator one

ColinSullivan1 (approved):
LGTM - Thanks!

PR #435, opened 2021-02-22
[added] PublicKeyFromSeed to obtain public key from seed

matthiashanel:
Signed-off-by: Matthias Hanel <mh@synadia.com>

when revising doc, I noticed that this is missing.


ColinSullivan1 (approved):
LGTM.  Thanks!

PR #436, opened 2021-02-25
Header updates

### LINKED ISSUE #419: Add newer header features
See:  https://github.com/nats-io/nats.go/commit/7ce305ab0b152eaaee7901d6062717786f20c528

ColinSullivan1:
Handle a description with the inline status.

Resolves #419


src/NATS.Client/Msg.cs
```diff
@@ -161,11 +180,37 @@ internal MsgHeader(byte[] bytes, int byteCount) : base()
...
+                            }
+
+                            // We can have code and description.
+
+                            // trim any leading whitespace.
+                            status = status.TrimStart();
+
+                            // check for description.
+                            int spaceIdx = status.IndexOf(' ');
+                            if (spaceIdx <= 0)
```
watfordgnf: This won't be 0 due to the TrimStart.
```suggestion
                            if (spaceIdx < 0)
```

watfordgnf (approved):
Minor suggestion.

src/NATS.Client/Msg.cs
ColinSullivan1: Thanks, sounds good.

PR #438, opened 2021-02-25
ISubscription should inherit IDisposable

### LINKED ISSUE #433: ISubscription should inherit IDisposable
From a user:

> Wouldn't it be better to make ISubscription inherit IDisposable rather than ISyncSubscription and IAsyncSubscription inherit both ISubscription and IDisposable individually? That way, if I want to maintain multiple subscriptions of both sync and async for a single connection where I keep all subscriptions as ISubscription in a dictionary keyed by subject, I have to cast appropriately to dispose all subscriptions when the connection close.

Would want to test to be sure an upgrade wouldn't break any code.

ColinSullivan1:
Resolves #433 

### DIFF
--- a/src/NATS.Client/IAsyncSubscription.cs
+++ b/src/NATS.Client/IAsyncSubscription.cs
@@ -12,9 +12,6 @@
 // limitations under the License.
 
 using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
 
 namespace NATS.Client
 {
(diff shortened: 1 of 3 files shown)


watfordgnf (approved):
Looks good, we'll probably want a note for folks who mock our subscriptions in testing that they may see this as a breaking change.

ColinSullivan1:
Thanks @watfordgnf for the reviews!

PR #439, opened 2021-02-25
Update Dependencies

### LINKED ISSUE #430: Update Dependencies (Security)
There are security findings in NATS.Client 0.11.0:

CVE-2017-0249
CVE-2019-0820
CVE-2018-8292

The two packages with the vulnerability are system.net.http and system.text.regularexpressions . These are transitive references that come in via netstandard.library.1.6.1.

https://www.whitesourcesoftware.com/vulnerability-database/CVE-2017-0249
https://www.whitesourcesoftware.com/vulnerability-database/CVE-2019-0820
https://www.whitesourcesoftware.com/vulnerability-database/CVE-2018-8292

ColinSullivan1:
Resolves #430 (as best we can).

### DIFF
--- a/src/Tests/IntegrationTests/TestBasic.cs
+++ b/src/Tests/IntegrationTests/TestBasic.cs
@@ -1965,7 +1965,6 @@ namespace IntegrationTests
             var exception = Record.Exception(() => Context.ConnectionFactory.CreateConnection());
 
             Assert.IsType<NATSConnectionException>(exception);
-            Assert.Equal("timeout", exception.Message);
         }
 
         /// <summary>
--- a/src/NATS.Client/NATS.Client.csproj
+++ b/src/NATS.Client/NATS.Client.csproj
@@ -39,5 +39,7 @@
 
   <ItemGroup>
     <PackageReference Include="System.Collections.Specialized" Version="4.3.0" />
+    <PackageReference Include="System.Net.Http" Version="4.3.4" />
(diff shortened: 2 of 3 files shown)


src/Tests/IntegrationTests/TestBasic.cs
```diff
@@ -1965,7 +1965,6 @@ public void TestServersOption()
             var exception = Record.Exception(() => Context.ConnectionFactory.CreateConnection());
 
             Assert.IsType<NATSConnectionException>(exception);
-            Assert.Equal("timeout", exception.Message);
```
ColinSullivan1, line 1968: This was flapping locally so removed the message check....

watfordgnf (approved):

src/Tests/IntegrationTests/TestBasic.cs
jasper-d: Thats strange. Do you remember what the actual message was?

ColinSullivan1: It was "Failed to connect", with an InnerException of ExtendedSocketConnection - "Connection refused...".  TBH, I'm OK with this, it's more accurate and the exception type didn't change.

jasper-d: I don't care about the message. I'm just a bit concerned that the behaviour isn't deterministic, but I cannot reproduce it locally.

PR #460, opened 2021-05-11
Heartbeat / rate limit / missed builder with

scottf:


src/NATS.Client/JetStream/ConsumerConfiguration.cs
```diff
@@ -32,6 +32,7 @@ public sealed class ConsumerConfiguration
         public string SampleFrequency { get; }
         public long RateLimit { get; }
         public long MaxAckPending { get; }
+        public Duration IdleHeartbeat { get; }
```
ColinSullivan1, line 35: Just a note, eventually these will require doc.

ColinSullivan1 (approved):
LGTM.

PR #461, opened 2021-05-12
Added test methods to create NATS server with JetStream enabled.

michaelkacher:
Added test methods to create NATS server with JetStream enabled.

@ColinSullivan1 Methods with a new name were created so there would not be ambiguity on nullable parameters if just the CreateFast(...) or CreateFastAndVerify(...) are invoked. Let me know your thoughts or what would be beneficial to be extended.

### DIFF
--- a/src/Tests/IntegrationTests/TestUtilities.cs
+++ b/src/Tests/IntegrationTests/TestUtilities.cs
@@ -72,6 +72,30 @@ namespace IntegrationTests
         public static NATSServer CreateFast(int port, string args = null)
             => new NATSServer(TimeSpan.Zero, port, args);
 
+        public static NATSServer CreateJetStreamFast(string args = null)
+            => CreateJetStreamFast(Defaults.Port, args);
+
+        public static NATSServer CreateJetStreamFast(int port, string args = null)
+        {
+            args = args == null
+                ? $"-js"
+                : $"{args} -js";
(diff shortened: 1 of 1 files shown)


src/Tests/IntegrationTests/TestUtilities.cs
```diff
@@ -72,6 +72,30 @@ public static NATSServer CreateFast(string args = null)
...
+        {
+            args = args == null
+                ? $"-js"
+                : $"{args} -js";
+
+            return new NATSServer(TimeSpan.Zero, port, args); 
+        }
+
+        public static NATSServer CreateJetStreamFastAndVerify(string args = null)
+            => CreateFastAndVerify(Defaults.Port, args);
```
ColinSullivan1: Should this be `CreateJetStreamFastAndVerify`?

michaelkacher: Yes, you are correct, fixed it! Would it be useful to use these methods to add some tests?

ColinSullivan1: It will be, but w haven't check in the code to start testing JetStream yet (it's in progress).  We'll use these for sure!  Thanks!

ColinSullivan1:
LGTM.  Thanks!

PR #464, opened 2021-06-06
Omit Base32 padding when encoding NKeys

### LINKED ISSUE #466: Using unsafe Random for Nkeys.CreateSeed()
While searching through the nats.net code base I noticed the method `Nkeys.CreateSeed()` is implemented like this:

```c#
private static string CreateSeed(byte prefixbyte) {
    byte[] rawSeed = new byte[32];

    Random rnd = new Random();
    rnd.NextBytes(rawSeed);

    return Encode(prefixbyte, true, rawSeed);
}
```

Shouldn't this code use `RandomNumberGenerator.GetBytes()` (or something similar) to generate a cryptographic secure seed? The `Random` class is not cryptographically secure (and [its value can even be predicted](https://manski.net/2016/08/breaking-nets-random-class/)).

sixlettervariables:
As seen on Slack, our .NET client apparently applies padding to Base32 encoded strings in line with the RFC but not in line with the other NATS clients. This removes the padding to match the Go client:
https://github.com/nats-io/nkeys/blob/44384414044304395985410ff55ed78cbf0b09a8/strkey.go#L52-L53

Also fixes #466.

### DIFF
--- a/src/NATS.Client/NATS.cs
+++ b/src/NATS.Client/NATS.cs
@@ -456,14 +456,14 @@ namespace NATS.Client
                 nextChar = (byte)((b << bitsRemaining) & 31);
             }
 
-            //if we didn't end with a full char
-            if (arrayIndex != charCount)
+            // if we didn't end with a full char
+            if (arrayIndex < charCount)
             {
                 returnArray[arrayIndex++] = ValueToChar(nextChar);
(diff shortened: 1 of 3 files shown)


ColinSullivan1 (approved):
LGTM.  Thanks!

PR #465, opened 2021-06-15
Implement the JetStream client API

ColinSullivan1:
This is a WIP.   Comments are welcome.


src/NATS.Client/JetStream/JetStream.cs
```diff
@@ -0,0 +1,245 @@
...
+            {
+                if (mh == null) mh = new MsgHeader();
+                mh.Set(MsgHeader.MsgIdHeader, opts.MessageId);
+            }
+
+            return mh;
+        }
+
+        private PublishAck PublishSync(string subject, byte[] data, PublishOptions opts)
+            => PublishSync(new Msg(subject, null, MergeHeaders(null, opts), data), opts);
```
watfordsuzy: This will end up calling `MergeHeaders` twice if `opts` is not null and has headers (via `PublishSync(Msg, PublishOptions)`). Probably best to just use `null` for headers here if it is being passed into `PublishSync` that will merge them.

src/NATS.Client/JetStream/JetStream.cs
```diff
@@ -0,0 +1,245 @@
...
+        public JetStreamOptions Options { get; }
+        public IConnection Connection { get; }
+        public int Timeout { get; }
+
+        private static readonly PublishOptions defaultPubOpts = PublishOptions.Builder().Build();
+
+        internal JetStream(IConnection connection, JetStreamOptions options)
+        {
+            Connection = connection;
+            Options = (options != null) ? options : JetStreamOptions.Builder().Build();
```
watfordsuzy:
```suggestion
            Options = options ?? JetStreamOptions.Builder().Build();
```

src/NATS.Client/JetStream/JetStream.cs
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        // Build the headers.  Take care not to unnecessarily allocate, we're
+        // in the fastpath here.
+        private MsgHeader MergeHeaders(MsgHeader headers, PublishOptions opts)
+        {
+            if (opts == null)
+                return null;
+
+            MsgHeader mh = headers != null ? headers : new MsgHeader();
```
watfordsuzy:
```suggestion
            MsgHeader mh = headers ?? new MsgHeader();
```

src/NATS.Client/JetStream/JetStream.cs
```diff
@@ -0,0 +1,245 @@
...
+
+            if (!string.IsNullOrEmpty(opts.ExpectedLastMsgId))
+            {
+                if (mh == null) mh = new MsgHeader();
```
watfordsuzy: `mh` is always non-null here (and the other checks) due to line 78 always setting it to a non-null value. These `null` checks should be elided.

src/NATS.Client/JetStream/JetStreamMsg.cs
```diff
@@ -0,0 +1,242 @@
...
+
+            int i;
+            for (i = 8; i < len; i++)
+            {
+                if (replySubj[i] == '.')
+                {
+                    i++;
+                }
+            }
+            return i == 7;
```
watfordsuzy: How would this be 7? Seems to start at 8 and only ever go up.

src/NATS.Client/JetStream/JetStreamMsg.cs
```diff
@@ -0,0 +1,242 @@
...
+            // TODO: Implement;
+            if ((sub is JetStreamSubscription) == false)
+            {
+                // Not a Jetstream enabled connection.
+                isPullMode = false;
+                return null;
+            }
+
+            isPullMode = (sub is JetStreamPullSubscription);
+            return ((JetStreamSubscription)sub).JS;
```
watfordsuzy: Could be simplified/.NETified some:
```suggestion
            if (sub is JetStreamSubscription jsSub)
            {
                isPullMode = jsSub is JetStreamPullSubscription;
                return jsSub.JS;
            }

            // Not a Jetstream enabled connection
            isPullMode = false;
            return null;
```

src/NATS.Client/JetStream/JetStreamMsg.cs
```diff
@@ -0,0 +1,242 @@
...
+    {
+        static readonly private byte[] ackAck = Encoding.ASCII.GetBytes("+ACK");
+	    static readonly private byte[] ackNak = Encoding.ASCII.GetBytes("-NAK");
+	    static readonly private byte[] ackProgress = Encoding.ASCII.GetBytes("+WPI");
+	    static readonly private byte[] ackTerm = Encoding.ASCII.GetBytes("+TERM");
+
+        /// <summary>
+        /// Gets the metadata associated with a JetStream message.
+        /// </summary>
+        public override MetaData Metadata { get; }
```
watfordsuzy: If the class is `MetaData` then I suggest the property matches:
```suggestion
        public override MetaData MetaData { get; }
```

src/NATS.Client/JetStream/JetStreamMsg.cs
```diff
@@ -0,0 +1,242 @@
...
+        /// Gets the consumer sequence number of the message.
+        /// </summary>
+        public Sequence Sequence { get; }
+
+
+        // Caller must ensure this is a JS message
+        internal MetaData(string metaData)
+        {
+            string[] tokens = metaData?.Split('.');
+            if (tokens.Length != 8)
```
watfordsuzy: `tokens` will be null if `metaData` is null:
```suggestion
            if (tokens?.Length != 8)
```

src/NATS.Client/JetStream/PushSubscribeOptions.cs
```diff
@@ -40,10 +27,17 @@ public sealed class PushSubscribeOptions
...
+            string deliverSubject) : base(stream, consumerConfiguration)
         {
-            _stream = stream;
-            _consumerConfiguration = consumerConfiguration;
+            string d = Validator.ValidateDurable(durable, false);
+            if (d == null && consumerConfiguration != null)
+            {
+                consumerConfiguration.Durable = Validator.ValidateDurable(consumerConfiguration.Durable, false);
+            }
+            consumerConfiguration.DeliverSubject = deliverSubject;
```
watfordsuzy: This is odd in a constructor where the only usage of `consumerConfiguration` is thru a base class; and depending on how it is used in the base class these statements may have no effect.

src/NATS.Client/Msg.cs
```diff
@@ -719,5 +741,68 @@ public bool HasHeaders
                 return header?.Count > 0;
             }
         }
+
+        #region JetStream
+
+        /// <summary>
+        /// Gets the metadata associated with a JetStream message.
+        /// </summary>
+        public virtual JetStream.MetaData Metadata { get { return null; } }
```
watfordsuzy: Same comment regarding the name of the property vs the class name.

src/NATS.Client/Msg.cs
```diff
@@ -450,7 +472,7 @@ public IEnumerator GetEnumerator()
     /// payload, optional header, and subscription information, sent or
     /// received by the client application.
     /// </summary>
-    public sealed class Msg
+    public class Msg
```
watfordsuzy, line 25: Perhaps instead of un-sealing message you could add the `IsJetStream` property plus have the `JetStreamMsg` class implement a new `IJetStreamMsg` interface that contains the methods you'd need. In JetStream code you'd be able to do:
```csharp
if (msg is IJetStreamMsg jsMsg)
{
    msg.Ack();
}
```

src/NATS.Client/JetStream/JetStream.cs
ColinSullivan1: Oh.. good catch.  Thanks!

src/NATS.Client/JetStream/JetStream.cs
ColinSullivan1: Thanks!  Been looking at too much java lately.  :)  I have a lot of changes in my branch so will make this change locally and push a bit later.

src/NATS.Client/JetStream/JetStream.cs
ColinSullivan1: Will fix...

src/NATS.Client/JetStream/JetStream.cs
ColinSullivan1: Will fix - thanks.

src/NATS.Client/JetStream/JetStreamMsg.cs
ColinSullivan1: Not sure what I was originally thinking here.  Will fix...

src/NATS.Client/JetStream/JetStreamMsg.cs
ColinSullivan1: +1

src/NATS.Client/JetStream/JetStreamMsg.cs
ColinSullivan1: Will fix...

src/NATS.Client/JetStream/JetStreamMsg.cs
scottf: Look at JsPrefixManager, it manages other prefixes as well Also I think the default is `$JS.API.`  The Java client calls JsPrefixManager.HasPrefix with the reply to to determine if it's a JS message

src/NATS.Client/JetStream/ApiResponse.cs
```diff
@@ -27,6 +27,14 @@ public abstract class ApiResponse
 
         internal ApiResponse(Msg msg) : this(Encoding.UTF8.GetString(msg.Data)) { }
```
AndyPook: this should have a null checks/guards
prob requires a little rearranging, maybe an `InitFormBytes` that can be used from the various ctors
means that the descendent types do not require the check

src/NATS.Client/JetStream/ConsumerConfiguration.cs
```diff
@@ -113,12 +121,12 @@ public static ConsumerConfigurationBuilder Builder(ConsumerConfiguration cc)
...
         {
             private DeliverPolicy _deliverPolicy = DeliverPolicy.All;
-            private AckPolicy _ackPolicy = JetStream.AckPolicy.Explicit;
+            private AckPolicy _ackPolicy = AckPolicy.Explicit;
             private ReplayPolicy _replayPolicy = ReplayPolicy.Instant;
             private string _durable;
             private string _deliverSubject;
             private long _startSeq;
-            private DateTime _startTime;
+            private DateTime _startTime = DateTime.MinValue;
```
AndyPook: not required.
`DateTime` is a value type, it defaults to `DateTime.MinValue`

src/NATS.Client/JetStream/JetStreamMsg.cs
```diff
@@ -0,0 +1,245 @@
...
+using System;
+using System.Text;
+
+namespace NATS.Client.JetStream
+{
+    public sealed class JetStreamMsg : Msg
+    {
+        static readonly private byte[] ackAck = Encoding.ASCII.GetBytes("+ACK");
+	    static readonly private byte[] ackNak = Encoding.ASCII.GetBytes("-NAK");
+	    static readonly private byte[] ackProgress = Encoding.ASCII.GetBytes("+WPI");
```
AndyPook: trivia: indent

src/NATS.Client/JetStream/JetStreamMsg.cs
```diff
@@ -0,0 +1,245 @@
...
+                return false;
+            }
+
+            // TODO - is this valid?
+            if (replySubj[0] != '$' ||
+                replySubj[1] != 'J' ||
+                replySubj[2] != 'S' ||
+                replySubj[3] != '.' ||
+                replySubj[4] != 'A' ||
+                replySubj[5] != 'C' ||
```
AndyPook: surely `if(!replySubj.StartsWith("$JS.ACK.", StringComparison.Ordinal))` would be more expressive (and the bcl prob has optimisations)

src/NATS.Client/JetStream/JetStreamMsg.cs
```diff
@@ -0,0 +1,245 @@
...
+                replySubj[6] != 'K' ||
+                replySubj[7] != '.')
+            {
+                return false;
+            }
+
+            int j = 0;
+            for (int i = 8; i < len; i++)
```
AndyPook: "$JS.ACK......" also passes (?)
does this need to "check" that there are actual tokens between the dots?

src/NATS.Client/Msg.cs
```diff
@@ -82,6 +82,26 @@ public sealed class MsgHeader : IEnumerable
...
+
+        /// <summary>
+        /// JetStream expected streame header name.
+        /// </summary>
+        public static readonly string ExpStreamHeader = "Nats-Expected-Stream";
+
+        /// <summary>
+        /// JetStream expected message ID header name.
+        /// </summary>
+        public static readonly string MsgIdHeader = "Nats-Msg-Id";
```
AndyPook: as these are `string`... should be `const` ?

AndyPook:
lots of places where _I_ would use `var`
is explicit typing a choice in this proj?

watfordsuzy:
My preference is the same as the .NET repo (sans Roslyn). Explicit types in all places except where obvious on the RHS (or the type is so terrible on the LHS that `var` improves readability).

[This project does not have a set-in-stone format and would not be considered idiomatic of modern .NET codebases.]

src/NATS.Client/JetStream/ConsumerConfiguration.cs
```diff
@@ -118,7 +126,7 @@ public sealed class ConsumerConfigurationBuilder
...
```
watfordsuzy: Why not a nullable DateTime?

src/NATS.Client/JetStream/JetStream.cs
```diff
@@ -75,34 +76,34 @@ private MsgHeader MergeHeaders(MsgHeader headers, PublishOptions opts)
...
-                if (mh == null) mh = new MsgHeader();
+                mh = headers ?? new MsgHeader();
                 mh.Set(MsgHeader.ExpLastIdHeader, opts.ExpectedLastMsgId);
             }
             if (opts.ExpectedLastSeq >= 0)
             {
-                if (mh == null) mh = new MsgHeader();
+                mh = headers ?? new MsgHeader();
                 mh.Set(MsgHeader.ExpLastSeqHeader, opts.ExpectedLastSeq.ToString());
             }
```
watfordsuzy: This pattern of `mh = headers ?? new MsgHeader();` is not actually correct as used here. You would need to do:
```csharp
            MsgHeader mh = headers;

            if (!string.IsNullOrEmpty(opts.ExpectedLastMsgId))
            {
                mh = mh ?? new MsgHeader();
                // mh ??= new MsgHeader(); // C# 8.0
                mh.Set(MsgHeader.ExpLastIdHeader, opts.ExpectedLastMsgId);
            }
```
Otherwise, if `headers` is null then each of these blocks is going to re-assign `mh`.

src/NATS.Client/JetStream/StreamConfiguration.cs
```diff
@@ -430,6 +435,11 @@ public StreamConfigurationBuilder(StreamConfiguration sc)
         /// </summary>
         /// <returns>The StreamConfiguration</returns>
         public StreamConfiguration Build() {
+            if (_subjects == null)
```
watfordsuzy:
```suggestion
            if (_subjects == null || !_subjects.Any())
```

AndyPook (commented):
> My preference is the same as the .NET repo (sans Roslyn). Explicit types in all places except where obvious on the RHS (or the type is so terrible on the LHS that `var` improves readability).
> 
> [This project does not have a set-in-stone format and would not be considered idiomatic of modern .NET codebases.]

_lots_ of places in dotnet repos where RHS is completely not obvious, but still uses `var` :)
lots of places in this repo like `List<string> results = new List<string>(values);` which is pretty "obvious". And lots of `var x = BuildThing()` which isn't

This is only the level of trivia (ie not important) and still "readable" so <shrug>
Just another "naming things" type discussion. Moving on...

src/NATS.Client/Internals/SimpleJson.cs
```diff
@@ -1012,6 +1012,10 @@ internal override void WriteToStringBuilder(StringBuilder aSB, int aIndent, int
                 aMode = JSONTextMode.Compact;
             foreach (var k in m_Dict)
             {
+                // Special NATS case: ignore null values or negative numbers.
+                if (k.Value.Value == null || k.Value.AsInt < 0)
```
scottf: Why are we ignoring negative numbers? We can't ignore negative numbers this way b/c 42000000000 nanos is a negative int

src/NATS.Client/JetStream/PushSubscribeOptions.cs
scottf: This has been changed. All validation is done in the builder since the constructor is private. Line 40 here also can NPE as one of the tests showed.

src/NATS.Client/JetStream/StreamConfiguration.cs
scottf: A subject is not required. If a stream is created and no subject is supplied, the server makes a subject with the same name as the stream.

src/NATS.Client/JetStream/ConsumerConfiguration.cs
scottf: I think it's ok to leave this. It's not obvious that it's a value type as a search on the internet shows that this question is asked a lot. Plus it's explicit as to what the default is.

src/NATS.Client/JetStream/JetStreamMsg.cs
scottf: All are developers converting tabs to spaces? Fixed

src/NATS.Client/JetStream/JetStreamMsg.cs
scottf: This is being replaced as it does not handle custom prefixes.

src/NATS.Client/JetStream/JetStreamMsg.cs
scottf: Being replaced...

src/NATS.Client/Msg.cs
scottf: changing to `public const`

src/NATS.Client/JetStream/JetStream.cs
scottf: Could also have been my code. I am trying to use do C# correctly.

src/NATS.Client/JetStream/JetStreamMsg.cs
scottf: fixed

src/NATS.Client/JetStream/JetStream.cs
scottf: fixed as suggested

src/NATS.Client/JetStream/ApiResponse.cs
scottf: so I built this, I'm a Java developer, so there may be some of that at work, sub classing and all... but this actually should never be called with a null message, so leaving it un guarded will suss out coding problems. I will give a check through the code to make sure...

src/Tests/UnitTests/TestMessageHeaders.cs
```diff
@@ -277,6 +285,76 @@ public void TestHeaderExceptions()
...
+            shouldNATSInvalidHeaderException("NATS/1.0\r\n");
+            shouldNATSInvalidHeaderException("NATS/1.0\r\n\r\n");
+            shouldNATSInvalidHeaderException("NATS/1.0\r\n\r\n\r\n");
+            shouldNATSInvalidHeaderException("NATS/1.0\r\nk1:v1");
+            shouldNATSInvalidHeaderException("NATS/1.0\r\nk1:v1\r\n");
+            shouldNATSInvalidHeaderException("NATS/1.0\r\nk1:v1\r\r\n");
+        }
+
+        private void shouldNATSInvalidHeaderException(string s)
+        {
```
sergeyshaykhullin, line 305: Should this be an ASCII?

src/NATS.Client/JetStream/ApiResponse.cs
AndyPook: it's a style choice I guess
The current way means that _every_ sub-class must do the null check. You just need one to forget.

"leaving it un guarded will suss out coding problems" ummm, ok. This is what guards are normally for. So that you get an meaningful exception in a useful place, rather than down in the guts of dotnet somewhere and then need to study the stack.

I thought java generally had the same style?

anyway, not super important :)

src/NATS.Client/JetStream/ConsumerConfiguration.cs
AndyPook: it is "ok", but a bit confusing to a dotnet geek :)
kinda the same as explicitly setting the `long` on the previous line to `0`. Or the `string`s to `null`.

Again, not "important", just looks a bit "odd"

src/Tests/UnitTests/TestMessageHeaders.cs
scottf: fixed

src/NATS.Client/JetStream/ConsumerConfiguration.cs
scottf: I definitely agree where privatives are concerned. Can other dotnet devs please chime in regarding DateTime?

src/NATS.Client/JetStream/ApiResponse.cs
scottf: So now that I think of it, this would never be called with a null message.  There is no need for a code guard here. These constructors are internal and it's expected that the developer know what they are doing. I would use a code guard when a valid value is null, but null is not valid here. On some languages, there are ways to mark parameters as not nullable. Is there the equivalent in .NET, I would use that I suppose

src/NATS.Client/JetStream/JetStream.cs
scottf: fixed

src/NATS.Client/JetStream/JetStreamMsg.cs
scottf: replaced by JsPrefixManager

src/NATS.Client/Msg.cs
scottf: fixed

src/NATS.Client/Msg.cs
scottf: There is an `IsJetStream` property now and the connection instantiates a `Msg` or a `JetStreamMsg` determined by using the `JsPrefixManager`. Do you still think we need the `IJetStreamMsg` interface since the property is there? Seems like we'd prefer to use a built in method instead of checking a class type, but that may be Java vs .NET so looking for more guidance here.

src/NATS.Client/JetStream/ConsumerConfiguration.cs
sixlettervariables: We would omit the MinValue assignment. 

We'd also prefer nullable if optional instead of relying on the default value which is less obvious.

src/NATS.Client/JetStream/ApiResponse.cs
sixlettervariables: Nullable reference types are available in the latest language version yes.

ColinSullivan1: > lots of places where _I_ would use `var`
> is explicit typing a choice in this proj?

It was simply a preference in initial development but I have no aversion to using var moving forward when it makes sense.

https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#implicitly-typed-local-variables

src/NATS.Client/Internals/SimpleJson.cs
ColinSullivan1: This was a bug... thanks.

src/NATS.Client/JetStream/Error.cs
```diff
@@ -33,18 +34,25 @@ internal Error(JSONNode node)
...
         }
 
         public override string ToString()
         {
-            if (string.IsNullOrEmpty(Desc)) {
-                return Code == NOT_SET
-                    ? "Unknown JetStream Error: " + _node.ToString()
-                    : "Unknown JetStream Error (" + Code + ")";
+            if (Code == NOT_SET) {
+                if (ApiErrorCode == NOT_SET) {
```
watfordsuzy:
```suggestion
            if (Code == NOT_SET)
            {
                if (ApiErrorCode == NOT_SET)
                {
```

src/NATS.Client/JetStream/Error.cs
```diff
@@ -33,18 +34,25 @@ internal Error(JSONNode node)
...
+                    return Desc;
+                }
+                return Desc + " [" + ApiErrorCode + "]";
             }
 
-            return Desc + " (" + Code + ")";
+            if (ApiErrorCode == NOT_SET) {
+                return Desc + " (" + Code + ")";
```
watfordsuzy:
```suggestion
            if (ApiErrorCode == NOT_SET)
            {
                return $"{Desc} ({Code})";
```

src/NATS.Client/JetStream/Error.cs
```diff
@@ -33,18 +34,25 @@ internal Error(JSONNode node)
...
+            }
+
+            return Desc + " (" + Code + ") [" + ApiErrorCode + "]";
```
watfordsuzy:
```suggestion
            return $"{Desc} ({Code}) [{ApiErrorCode}]";
```

src/NATS.Client/JetStream/Error.cs
```diff
@@ -33,18 +34,25 @@ internal Error(JSONNode node)
...
         {
-            if (string.IsNullOrEmpty(Desc)) {
-                return Code == NOT_SET
-                    ? "Unknown JetStream Error: " + _node.ToString()
-                    : "Unknown JetStream Error (" + Code + ")";
+            if (Code == NOT_SET) {
+                if (ApiErrorCode == NOT_SET) {
+                    return Desc;
+                }
+                return Desc + " [" + ApiErrorCode + "]";
```
watfordsuzy:
```suggestion
                return $"{Desc} [{ApiErrorCode}]";
```

watfordsuzy (commented):
Small stuff, places with Java style braces should be reformatted but I only tagged the first one. Perhaps since this is a port of the Java client we can wait for a bit and reformat it all once the bulk of it is in place.

src/NATS.Client/JetStream/Error.cs
scottf: Just force of habit on the brackets, I'll fix even though it's clearly not as good! :) Also I will use the interpolation, that's much better than the concatenation

src/NATS.Client/JetStream/Error.cs
scottf: will fix

src/NATS.Client/JetStream/Error.cs
scottf: will fix

src/NATS.Client/JetStream/Error.cs
scottf: will fix

src/NATS.Client/JetStream/Error.cs
watfordsuzy: I do a lot of open source Java on the side and I rely heavily on VS Code's reformatting to make it more Java like...so I definitely get it!

src/NATS.Client/Internals/JsonUtils.cs
```diff
@@ -73,5 +74,10 @@ internal static JSONArray ToArray(List<string> list)
             }
             return arr;
         }
+ 
+        public static byte[] SimpleMessageBody(string name, object value)
+        {
+            return Encoding.ASCII.GetBytes("{\"" + name + "\":" + value.ToString() + "}");
```
watfordsuzy: Also, will we need to consider escaping of `value` here, or is that on the caller?
```suggestion
            return Encoding.UTF8.GetBytes("{\"" + name + "\":" + value.ToString() + "}");
```

watfordsuzy (commented):
One note about JSON encoding.

src/NATS.Client/Internals/JsonUtils.cs
scottf: right now the only use is where the object is a number. Also internal uses should never have anything other than ascii. Also I was going to use interpolation, but didn't know what how to properly handle the double quotes.

watfordsuzy: Suggested as JSON is UTF8 encoded. I didn't suggest interpolation due to the braces, otherwise the rules for escaping quotes are based on whether or not the string is literal or not (i.e. prefixed with `@`):
```suggestion
            return Encoding.UTF8.GetBytes($"{{\"{name}\":{value}}}");
```
An alternative using an interpolated literal string:
```csharp
            return Encoding.UTF8.GetBytes($@"{{""{name}"":{value}}}");
```

src/NATS.Client/Internals/JsonUtils.cs
```diff
@@ -75,9 +75,14 @@ internal static JSONArray ToArray(List<string> list)
             return arr;
         }
  
-        public static byte[] SimpleMessageBody(string name, object value)
+        public static byte[] SimpleMessageBody(string name, long value)
         {
-            return Encoding.ASCII.GetBytes("{\"" + name + "\":" + value.ToString() + "}");
+            return Encoding.ASCII.GetBytes("{\"" + name + "\":" + value + "}");
+        }
```
sixlettervariables: We need to be careful marshalling `long` to and from JSON as we only basically have 53 bits.

scottf: Can you clarify what you mean? 53 or 63 bits? SimpleMessageBody is just a convenience method. Maybe it should be `internal`? Is ToString redundant? The current use is for a message sequence number, which in go is a uint64. I'll try to provide an overload.

sixlettervariables: JSON cannot marshal integer values beyond 53-bits because JavaScript lacks an integer type all together (it effectively uses `double`). We need to send these as strings if they are in fact UInt64.

Examples of problems:
- https://randomascii.wordpress.com/2020/09/27/floating-point-in-the-browser-part-1-impossible-expectations/

scottf: Our schema expects a number. I'll verify that the .NET client handles it fine. I don't think it's our problem if javascript in browsers can't handle it. Also 53 bits is a pretty huge number so I think we'll be okay. The java client currently only supports 32 bits (signed long) and we have not heard any issues. I'm fixing that anyway though.

src/NATS.Client/JetStream/IJetStreamSubscription.cs
```diff
@@ -93,6 +93,6 @@ public interface IJetStreamPullSubscription : IJetStreamSubscription, ISyncSubsc
         /// <param name="batchSize">the size of the batch</param>
         /// <param name="maxWaitMillis">The maximum time to wait for the first message.</param>
         /// <returns></returns>
-        JetStreamMsg[] Fetch(int batchSize, long maxWaitMillis);
+        List<Msg> Fetch(int batchSize, int maxWaitMillis);
```
watfordsuzy:
```suggestion
        IList<Msg> Fetch(int batchSize, int maxWaitMillis);
```

scottf: got it. List is concrete like ArrayList in java, IList is the interface like List is in Java. Fixed

src/NATS.Client/JetStream/JetStream.cs
```diff
@@ -88,6 +89,24 @@ private PublishAck PublishSyncInternal(string subject, byte[] data, MsgHeader hd
...
+            MsgHeader merged = MergePublishOptions(hdr, options);
+            Msg msg = new Msg(subject, null, merged, data);
+
+            if (JetStreamOptions.IsPublishNoAck)
+            {
+                Conn.Publish(msg);
+                return null;
+            }
+
+            Duration timeout = options == null ? JetStreamOptions.RequestTimeout : options.StreamTimeout;
```
watfordsuzy, line 99: We'll want a way to do infinite timeout, but with the `timeout.Millis` call below performing math I'm not sure that's as simple anymore.

src/NATS.Client/JetStream/JetStream.cs
```diff
@@ -88,6 +89,24 @@ private PublishAck PublishSyncInternal(string subject, byte[] data, MsgHeader hd
...
+
+            return Conn.RequestAsync(msg, timeout.Millis)
+                .ContinueWith(async antecedent => ProcessPublishResponse(antecedent.Result, options), CancellationToken.None)
```
watfordsuzy, line 102: Drop the `async` if you are not using `await`:
```suggestion
                .ContinueWith(antecedent => ProcessPublishResponse(antecedent.Result, options), CancellationToken.None)
```

watfordsuzy: On second thought, is there a reason that `PublishAsyncInternal` cannot be declared `async` and you `await Conn.RequestAsync(...)`?
```csharp
        private async Task<PublishAck> PublishAsyncInternal(string subject, byte[] data, MsgHeader hdr, PublishOptions options)
        {
            MsgHeader merged = MergePublishOptions(hdr, options);
            Msg msg = new Msg(subject, null, merged, data);

            if (JetStreamOptions.IsPublishNoAck)
            {
                await Conn.PublishAsync(msg);
                return null;
            }

            Duration timeout = options == null ? JetStreamOptions.RequestTimeout : options.StreamTimeout;

            var result = await Conn.RequestAsync(msg, timeout.Millis);

            return ProcessPublishResponse(result, options);
        }
```

scottf: Conn.RequestAsync already returns a future. It hides the async under the covers. I understood that antecedent just tacks itself on to the task. I copied from here: https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/chaining-tasks-by-using-continuation-tasks and certainly may not have used it correctly, but it functionally works.

src/NATS.Client/JetStream/JetStream.cs
scottf: The timeout applies to the publish portion and is required.

src/NATS.Client/JetStream/JetStream.cs
watfordsuzy: In general you don't really want to use ContinueWith if it can be avoided. Lots of little caveats here and there as to how the task is attached, cancellation, exceptions; best to use `async` + `await` if you're not just immediately chaining a result. It can also get you into trouble if at any point you mix sync-over-async and async code.

scottf: Adding an issue so we don't lose track of this. https://github.com/nats-io/nats.net/issues/475

PR #485, opened 2021-08-23
Proposal for missing implementation of Jetstream.PublishAsyncInternal

kleferbe:
I just came across this and I've probably overlooked something but this would be the simplest implementation I can thing of. Perhaps you can make some use of it. If not - never mind.

### DIFF
--- a/src/NATS.Client/JetStream/JetStream.cs
+++ b/src/NATS.Client/JetStream/JetStream.cs
@@ -84,7 +84,7 @@ namespace NATS.Client.JetStream
             return ProcessPublishResponse(Conn.Request(msg), options);
         }
 
-        private Task<PublishAck> PublishAsyncInternal(string subject, byte[] data, MsgHeader hdr, PublishOptions options)
+        private async Task<PublishAck> PublishAsyncInternal(string subject, byte[] data, MsgHeader hdr, PublishOptions options)
         {
             MsgHeader merged = MergePublishOptions(hdr, options);
             Msg msg = new Msg(subject, null, merged, data);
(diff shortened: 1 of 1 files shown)


src/NATS.Client/JetStream/JetStream.cs
```diff
@@ -97,34 +97,8 @@ private Task<PublishAck> PublishAsyncInternal(string subject, byte[] data, MsgHe
...
-            // {
-            //     Task<Msg> msgTask = Conn.RequestAsync(msg, timeout.Millis);
-            //     msgTask.Wait();
-            //     return ProcessPublishResponse(msgTask.Result, options);
-            // });
-            //
-            // return await paTask.ConfigureAwait(false);
-            // -----
+            var result = await Conn.RequestAsync(msg, timeout.Millis).ConfigureAwait(false);
+            return ProcessPublishResponse(result, options);
```
scottf, line 101: Does making the method async automatically make a Task of the return type and return without blocking? This seems too easy, but if it does the trick I'm good with it!

kleferbe: yes, exactly. The returned Task represents the return value that might or might not be ready at the time the method returns. (In Javascript/Typescript async/await works the same but the Task is named Promise there).

When your method is completely synchronous it will still work but you will get a compiler warning that recommends to remove async (and the Task wrapping). So you have to await something for the compiler to be happy, but not in all code paths.

sixlettervariables: @scottf yes, the simple answer is as long as you `await` something you'll be "asynchronous" (in some fashion).

sixlettervariables (approved):
Approving without CI as we do not have any guarantees around JS for now. We'll want to confirm our code is okay with `.ConfigureAwait(false)` (we should be, but we do lots of async-over-sync).

PR #489, opened 2021-08-25
ack meta last minute clarification

scottf:


sixlettervariables:
CI is failing with a build error, can you check that out?

PR #498, opened 2021-10-10
Parity changes

scottf:
Parity changes found in Architecture Repo Issues including subscription work and auto status management.
https://github.com/nats-io/nats-architecture-and-design/issues

Testing and Samples tracked for parity with Java client here:
https://docs.google.com/spreadsheets/d/1kM_z8oq1EFV53Wau4A1BwlqP9PKYTNdThj8oN8qzX2I/edit?usp=sharing


src/NATS.Client/AckType.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        public static AckType AckAck = new AckType("+ACK");
+        public static AckType AckNak = new AckType("-NAK");
+        public static AckType AckProgress = new AckType("+WPI");
+        public static AckType AckTerm = new AckType("+TERM");
+
+        public string Text { get; }
+        public byte[] Bytes { get; }
+
+        public AckType(string text)
```
ColinSullivan1, line 28: Does this need to be public?

src/NATS.Client/Conn.cs
```diff
@@ -747,8 +756,15 @@ internal Connection(Options options)
             callbackScheduler.Start();
 
             globalRequestInbox = NewInbox();
+
+            beforeQueueProcessor = msg => msg;
         }
 
+        internal void SetBeforeQueueProcessor(Func<Msg, Msg> beforeQueueProcessor)
+        {
+            this.beforeQueueProcessor = beforeQueueProcessor;
```
ColinSullivan1: Nitpick, could this be a property?

src/NATS.Client/Conn.cs
```diff
@@ -1139,9 +1173,8 @@ internal bool connect(Srv s, out Exception exToThrow)
                         if (!ex.IsAuthorizationViolationError() && !ex.IsAuthenticationExpiredError())
                             throw;
 
-                        var aseh = opts.AsyncErrorEventHandler;
-                        if (aseh != null)
-                            callbackScheduler.Add(() => aseh(s, new ErrEventArgs(this, null, ex.Message)));
+                        // handler is defaulted in constructor
+                        callbackScheduler.Add(() => opts.AsyncErrorEventHandler(s, new ErrEventArgs(this, null, ex.Message)));
```
ColinSullivan1: What happens if the user explicitly sets the error handler to null?

ColinSullivan1: e.g. `opts.AsyncErrorEventHandler = null;`

scottf: when the connection is made, the event handlers are checked. If the user changes the handler after the connection, I suppose this would fail. Frankly I don't see why there is a builder and the ability to just set properties directly. I would redo the option to make them immutable, instead of just public properties.

src/NATS.Client/Conn.cs
scottf: yes, Probably a carry over from Java

src/NATS.Client/AckType.cs
scottf: can't test them without being public since test classes are in another assembly, unless there is something .NET I'm missing

ColinSullivan1: Could unit test...

src/NATS.Client/Conn.cs
ColinSullivan1: It's the pattern for event handlers in .NET.

src/NATS.Client/Internals/JetStreamConstants.cs
```diff
@@ -126,16 +126,22 @@ public static class JetStreamConstants
...
         /// JetStream expected last message ID header Name.
         /// </summary>
         public const string ExpLastIdHeader = "Nats-Expected-Last-Msg-Id";
+        
+        /// <summary>
+        /// JetStream expected last sequence header name.
+        /// </summary>
+        public const string ExpLastSubjectSeqHeader = "Nats-Expected-Last-Subject-Sequence";
+        
+        public const string LastConsumerHdr = "Nats-Last-Consumer";
```
ColinSullivan1, line 140: Should these have header spelled out like the one above?  Also could use some comments.

ColinSullivan1 (approved):
LGTM.  Thanks!  Could you fix the constants with "Header" in them before merging?

src/NATS.Client/Conn.cs
scottf: fixed

src/NATS.Client/Conn.cs
scottf: fixed

PR #503, opened 2021-11-01
Issue 502

scottf:
The comparison of a default consumer often used for subscribing to pre-made consumers was failing indicating a change was requested, when it was not. This problem is fixed with this PR


src/NATS.Client/Internals/JsonUtils.cs
```diff
@@ -126,5 +128,49 @@ public static byte[] Serialize(JSONNode node)
...
+            DateTime dt = value.GetValueOrDefault(DateTime.MinValue);
+            if (!dt.Equals(DateTime.MinValue))
+            {
+                o[field] = UnsafeToString(dt);
+            }
+        }
+
+        public static void AddField(JSONObject o, string field, long? value)
+        {
+            long l = value.GetValueOrDefault(-1);
```
sixlettervariables: Are these values never actually zero?

scottf: -1 is often a default value. 0 is valid with meaning most of the time.

sixlettervariables: Why not just test for null as these are nullable rather than use a sentinel value?

scottf: It was late and my editor was giving complaining so I just switched to something the was working. I guess resharper just needed a nap, it's fixed now.

src/NATS.Client/Internals/JsonUtils.cs
```diff
@@ -126,5 +133,53 @@ public static byte[] Serialize(JSONNode node)
...
+        {
+            if (value != null)
+            {
+                o[field] = UnsafeToString(value.Value);
+            }
+        }
+
+        public static void AddField(JSONObject o, string field, long? value)
+        {
+            if (value != null && value >= 0)
```
sixlettervariables, line 155: Double checking that we should never serialize negative values? Seems off, but if that's the protocol I'm okay.

scottf: -1 often represents default and there is no need to write it across the wire. Negative values for settings are not valid, at least for all current fields.

src/NATS.Client/Internals/JsonUtils.cs
```diff
@@ -126,5 +133,53 @@ public static byte[] Serialize(JSONNode node)
                     return o.ToString();
             }
         }
+
+        public static void AddField(JSONObject o, string field, string value)
```
ColinSullivan1: Should these be internal?

documentation/DoxyFile.NATS.Client
```diff
@@ -38,7 +38,7 @@ PROJECT_NAME           = "NATS .NET Client"
 # could be handy for archiving the generated documentation or if some version
 # control system is used.
 
-PROJECT_NUMBER         = 0.14.0
+PROJECT_NUMBER         = 0.14.1
```
ColinSullivan1, line 41: 1.0.0?

ColinSullivan1: Or will this be an interim 0.14.1 release?

src/NATS.Client/JetStream/ConsumerConfiguration.cs
```diff
@@ -299,7 +307,7 @@ public ConsumerConfigurationBuilder WithDeliverGroup(string group)
             /// </summary>
             /// <param name="sequence">the start sequence</param>
             /// <returns>The ConsumerConfigurationBuilder</returns>
-            public ConsumerConfigurationBuilder WithStartSequence(ulong sequence)
+            public ConsumerConfigurationBuilder WithStartSequence(ulong? sequence)
```
ColinSullivan1, line 310: This is nullable to be able to unset a sequence?

src/NATS.Client/Internals/JsonUtils.cs
scottf: fixing

src/NATS.Client/JetStream/ConsumerConfiguration.cs
scottf: This is nullable so I can properly know when they do not even set a value, so comparisons of the "default" config is not a change versus one that is on the server. When creating a consumer, null is considered "unset"

ColinSullivan1 (approved):
LGTM.

PR #523, opened 2021-12-05
Key Value

scottf:


src/NATS.Client/KeyValue/KeyValue.cs
```diff
@@ -0,0 +1,198 @@
...
+                int timeout = 5000; // give plenty of time for the first
+                while (true)
+                {
+                    Msg m = sub.NextMessage(timeout);
+                    action.Invoke(m);
+                    timeout = 100; // the rest should come pretty quick
+                }
+            }
+            catch (NATSTimeoutException) { /* no more messages */ }
+            sub.Unsubscribe();
```
sixlettervariables: `sub.Unsubscribe` should be in a finally block.

src/NATS.Client/KeyValue/KeyValue.cs
```diff
@@ -0,0 +1,198 @@
...
+                .Build();
+            IJetStreamPushSyncSubscription sub = js.PushSubscribeSync(subject, pso);
+            try
+            {
+                int timeout = 5000; // give plenty of time for the first
+                while (true)
+                {
+                    Msg m = sub.NextMessage(timeout);
+                    action.Invoke(m);
+                    timeout = 100; // the rest should come pretty quick
```
sixlettervariables: These timeout values are curiously fixed. Is timing out truly our only means of detecting we've hit the end?

src/NATS.Client/KeyValue/KeyValueWatchSubscription.cs
```diff
@@ -0,0 +1,93 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using NATS.Client.Internals;
+using NATS.Client.JetStream;
+
+namespace NATS.Client.KeyValue
+{
+    public class KeyValueWatchSubscription
```
sixlettervariables: Consider making this IDisposable and having Dispose call Unsubscribe.

src/NATS.Client/KeyValue/KeyValue.cs
scottf: fixing

src/NATS.Client/KeyValue/KeyValue.cs
scottf: Adding check for Pending == 0

src/NATS.Client/KeyValue/KeyValueWatchSubscription.cs
scottf: done

aricart (approved):
LGTM

PR #531, opened 2021-12-23
kv subject prefix changes to properly support accounts

scottf:


src/NATS.Client/Internals/Validator.cs
```diff
@@ -543,5 +543,21 @@ public static Duration EnsureDurationNotLessThanMin(long providedMillis, Duratio
         {
             return EnsureNotNullAndNotLessThanMin(Duration.OfMillis(providedMillis), minimum, dflt);
         }
+         
+        public static bool Equal(byte[] a, byte[] a2) {
```
sixlettervariables, line 547: `a.SequenceEqual(b)` would be preferred.

scottf: Either or both a and a2 can be null. SequenceEqual throws on nulls.

sixlettervariables: Past the null checks I meant, was reviewing on my phone.

scottf: I initially considered and looked at SequenceEqual code. Its generic, it instantiates a comparer object, checks for nulls to throw, checks for a known collection type, so much just to compare a simple byte array? This code is pretty simple, I think it's ok, but if it's absolutely have to be done I will change it.

aricart (approved):
LGTM

PR #539, opened 2022-01-18
Implement Async Get/Delete

thinkbeforecoding:
This PR implement Task based async for GetMessage/DeleteMessage.

I have an extra question on this. Would it be ok to also implement GetMessage on IJetStream ? It would return a Msg instead of MsgInfo in that case.

https://github.com/nats-io/nats.net/issues/540

### DIFF
--- a/src/NATS.Client/JetStream/IJetStreamManagement.cs
+++ b/src/NATS.Client/JetStream/IJetStreamManagement.cs
@@ -12,6 +12,7 @@
 // limitations under the License.
 
 using System.Collections.Generic;
+using System.Threading.Tasks;
 
 namespace NATS.Client.JetStream
 {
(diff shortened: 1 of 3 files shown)


sixlettervariables (dismissed):

sixlettervariables:
@thinkbeforecoding can you open an issue to track your question regarding the other method family.

scottf:
@thinkbeforecoding a couple things. Please sign your commits, this is the policy for our repos. 

Regarding the change, you could always do it yourself by wrapping the sync call. The problem with adding this to the api is that we have to consider it in all clients, not that all clients have to be exactly equal but we do try to maintain some amount of parity. And since you can wrap the call yourself and the management context is actually very light, I don't see the problem with using both the management api and the regular api. We are really just separating concerns by having different interfaces.

scottf:
Also GetMessage is a JS API call, which is why it returns MessageInfo instead of a Message, which is returned in subscriptions.

sixlettervariables:
In .NET you should avoid wrapping sync in async calls whenever possible. "Turtles all the way down" is best, so to speak.

thinkbeforecoding:
I'm ok to leave GetMessage in the management api. But having an Async version (returning a task) is very important for .Net performance.

thinkbeforecoding:
I signed the commit 👍

scottf:
This will be addressed in V2

JohannesEH:
@scottf Any more details about v2? Do you have an idea of how far down the line it is? Like this year, next year? Is it a complete rewrite?

scottf:
See V2

PR #544, opened 2022-01-25
Added netstandard2.0 target

JohannesEH:
Added a "netstandard2.0" target to NATS.Client.csproj to simplify the transitive dependency hierarchy for modern projects.

### DIFF
--- a/src/NATS.Client/NATS.Client.csproj
+++ b/src/NATS.Client/NATS.Client.csproj
@@ -1,8 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks Condition="$(OS) == 'Windows_NT'">netstandard1.6;net46</TargetFrameworks>
-    <TargetFramework Condition="$(OS) != 'Windows_NT'">netstandard1.6</TargetFramework>
+    <TargetFrameworks Condition="$(OS) == 'Windows_NT'">netstandard1.6;netstandard2.0;net46</TargetFrameworks>
+    <TargetFramework Condition="$(OS) != 'Windows_NT'">netstandard1.6;netstandard2.0</TargetFramework>
     <Title>NATS .NET Client</Title>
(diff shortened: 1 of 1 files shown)


sixlettervariables (approved):

thinkbeforecoding (commented):
Thank you for this change that will simplify dependencies.
There are some minor changes that could make it even simpler.

src/NATS.Client/NATS.Client.csproj
```diff
@@ -1,8 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks Condition="$(OS) == 'Windows_NT'">netstandard1.6;net46</TargetFrameworks>
-    <TargetFramework Condition="$(OS) != 'Windows_NT'">netstandard1.6</TargetFramework>
+    <TargetFrameworks Condition="$(OS) == 'Windows_NT'">netstandard1.6;netstandard2.0;net46</TargetFrameworks>
```
thinkbeforecoding, line 4: You can actually build the net46 target on non-windows OS using the [Microsoft.NETFramework.ReferenceAssemblies](https://www.nuget.org/packages/Microsoft.NETFramework.ReferenceAssemblies/) nuget.

It would of course be usable at runtime only on windows, but would let the nuget be compiled on Linux/MacOs

JohannesEH: Thanks. :) I guess that you are right. It would get rid of the condition on `<TargetFrameworks ...>` and simplify the file a bit. However it doesn't relate to my change and the `netstandard2.0` addition (more related to the existing `net46` target), so shouldn't it be in a different PR?

sixlettervariables: Yes a second PR preferably.

thinkbeforecoding: Yes! Perfect, thx !

JohannesEH:
@danielwertheim what do you think about this?

PR #548, opened 2022-02-01
Fix the default reconnect delay

### LINKED ISSUE #547: Application doesn't attempt to reconnect.
#### Versions of `NATS.Client` and `nats-server`:
Version 0.14.0

#### OS/Container environment:
Windows

#### Steps or code to reproduce the issue:

The unit test should pass

#### Expected result:

The application reconnects

#### Actual result:

The application fails to reconnect

```
[Fact]
public void TestReconnectTiming()
{
    var closedEv = new AutoResetEvent(false);
    var disconnEv = new AutoResetEvent(false);

    var opts = Context.GetTestOptionsWithDefaultTimeout(Context.Server1.Port);
    opts.MaxReconnect = 3;
    opts.ReconnectWait = 1000;
    opts.SetReconnectJitter(0, 0);

    Stopwatch sw = new Stopwatch();

    opts.DisconnectedEventHandler = (obj, args) =>
    {
        sw.Start();
    };
    opts.ClosedEventHandler = (obj, args) =>
    {
        closedEv.Set();
        sw.Stop();
    };

    using (var s = NATSServer.CreateFastAndVerify(Context.Server1.Port))
    {
        using (var c = Context.ConnectionFactory.CreateConnection(opts))
        {
            // shutdown the server.
            s.Shutdown();

            int ExpectedMin = opts.MaxReconnect * opts.ReconnectWait;
            // Make until we're closed, add some slack.
            Assert.True(closedEv.WaitOne(ExpectedMin + 1000));
            Assert.True(sw.ElapsedMilliseconds > (opts.MaxReconnect * opts.ReconnectWait),
                $"Elapsed {sw.ElapsedMilliseconds} ms < expected Minimum {ExpectedMin} ms");
        }
    }
}
```

ColinSullivan1:
The original default reconnect delay handler was overridden.  Revert back to using the original.

Resolves #547  

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -722,6 +722,11 @@ namespace NATS.Client
         {
             opts = new Options(options);
 
+            if (opts.ReconnectDelayHandler == null)
+            {
+                opts.ReconnectDelayHandler = DefaultReconnectDelayHandler;
+            }
+
             PING_P_BYTES = Encoding.UTF8.GetBytes(IC.pingProto);
             PING_P_BYTES_LEN = PING_P_BYTES.Length;
 
@@ -1620,7 +1625,7 @@ namespace NATS.Client
                             // If unset, the default handler will be called which uses an
                             // auto reset event to wait, unless kicked out of a close
                             // call.
-                            opts?.ReconnectDelayHandler(this, new ReconnectDelayEventArgs(wlf - 1));
+                            opts.ReconnectDelayHandler(this, new ReconnectDelayEventArgs(wlf - 1));
                         }
                         catch { } // swallow user exceptions
                     }
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -1622,10 +1622,7 @@ namespace NATS.Client
                     {
                         try
                         {
-                            // If unset, the default handler will be called which uses an
-                            // auto reset event to wait, unless kicked out of a close
-                            // call.
-                            opts.ReconnectDelayHandler(this, new ReconnectDelayEventArgs(wlf - 1));
+                            opts?.ReconnectDelayHandler(this, new ReconnectDelayEventArgs(wlf - 1));
                         }
                         catch { } // swallow user exceptions
                     }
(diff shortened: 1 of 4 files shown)


src/NATS.Client/Conn.cs
```diff
@@ -1620,7 +1625,7 @@ private void doReconnect()
                             // If unset, the default handler will be called which uses an
                             // auto reset event to wait, unless kicked out of a close
                             // call.
-                            opts?.ReconnectDelayHandler(this, new ReconnectDelayEventArgs(wlf - 1));
+                            opts.ReconnectDelayHandler(this, new ReconnectDelayEventArgs(wlf - 1));
```
scottf: I guess opts could have been null at one point?

src/NATS.Client/Conn.cs
```diff
@@ -722,6 +722,11 @@ internal Connection(Options options)
         {
             opts = new Options(options);
 
+            if (opts.ReconnectDelayHandler == null)
+            {
+                opts.ReconnectDelayHandler = DefaultReconnectDelayHandler;
+            }
+
```
scottf, line 729: Yes, just put the original code back

src/NATS.Client/Options.cs
```diff
@@ -49,35 +49,35 @@ public sealed class Options
         /// Represents the method that will handle an event raised 
         /// when a connection is closed.
         /// </summary>
-        public EventHandler<ConnEventArgs> ClosedEventHandler;
+        public EventHandler<ConnEventArgs> ClosedEventHandler = null;
```
scottf: This doesn't hurt, but the object should be null if not assigned. The ide marked these as "Initializing field by default value is redundant"

scottf (approved):

src/NATS.Client/Options.cs
sixlettervariables: This is redundant and can be removed.

ColinSullivan1: Will remove...

src/NATS.Client/Conn.cs
ColinSullivan1: I don't see how it could, tbh.

ColinSullivan1: Will add it back just in case.

ColinSullivan1: Fixed...

PR #550, opened 2022-02-04
Fix push subscribe async in bound mode

kleferbe:
All `JetStream` subscribe methods should allow an empty subject if the given `SubscribeOptions.Bind` is `true`. Otherwise it would not be possible to subscribe to an existing consumer that has an empty filter subject.


scottf (approved):
LGTM

scottf:
Using administrator privileges to merge with unsigned commit.

PR #562, opened 2022-02-16
Kv parity issues 94, 95, 96

scottf:
https://github.com/nats-io/nats-architecture-and-design/issues/94 Items 1 and 3
https://github.com/nats-io/nats-architecture-and-design/issues/95
https://github.com/nats-io/nats-architecture-and-design/issues/96

* 94: v2.7.2+, the KV stream need to be created with the DiscardNew
* 94: use Pending+Delivered.Consumer from the subscription to determine if end of data marker should be sent
* 95: use of prefix as defined in https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-19.md
* 96 PurgeDeletes should keep "recent" markers


src/Tests/UnitTests/JetStream/TestServerInfo.cs
```diff
@@ -46,6 +46,31 @@ public void JsonIsReadProperly()
...
+        [Fact]
+        public void ServerVersionComparisonsWork()
+        {
+            ServerInfo info = new ServerInfo(_json);
+            ServerInfo info234 = new ServerInfo(_json.Replace("1.2.3", "2.3.4"));
+            ServerInfo info235 = new ServerInfo(_json.Replace("1.2.3", "2.3.5"));
+            ServerInfo info235Beta2 = new ServerInfo(_json.Replace("1.2.3", "2.3.5-beta.2"));
+            Assert.True(info.IsOlderThanVersion("2.3.4"));
+            Assert.True(info234.IsOlderThanVersion("2.3.5"));
+            Assert.True(info235.IsOlderThanVersion("2.3.5-beta.2"));
```
caleblloyd: This is false - `2.3.5` is newer than `2.3.5-<pre>`.  But do we even need this level of granularity?  If we are trying to check features against server versions then we probably want to consider `2.3.5-beta.2` to have the same features as `2.3.5`

src/Tests/UnitTests/JetStream/TestServerInfo.cs
```diff
@@ -46,6 +46,31 @@ public void JsonIsReadProperly()
...
+            Assert.True(info.IsSameVersion("1.2.3"));
+            Assert.True(info234.IsSameVersion("2.3.4"));
+            Assert.True(info235.IsSameVersion("2.3.5"));
+            Assert.True(info235Beta2.IsSameVersion("2.3.5-beta.2"));
+            Assert.False(info235.IsSameVersion("2.3.4"));
+            Assert.False(info235Beta2.IsSameVersion("2.3.5"));
+            Assert.True(info234.IsNewerVersionThan("1.2.3"));
+            Assert.True(info235.IsNewerVersionThan("2.3.4"));
+            Assert.True(info235Beta2.IsNewerVersionThan("2.3.5"));
```
caleblloyd: same issue - `2.3.5` is newer than `2.3.5-<pre>`

src/NATS.Client/JetStream/ServerInfo.cs
```diff
@@ -65,5 +66,29 @@ public ServerInfo(string json)
                 .Select(n => n.Value)
                 .ToArray();
         }
+
+        private string GetComparableVersion(string vString) {
```
caleblloyd, line 70: Consider using the [Version Class](https://docs.microsoft.com/en-us/dotnet/api/system.version?view=net-6.0#properties) for this comparison.  It is not 1:1 with SemVer, but do we even need to consider `prerelease` data?  Or can we just get away with comparing Major/Minor/Patch?  Example:

```c#
using System;
using System.Text.RegularExpressions;

public class Example
{
   public static void Main()
   {
      Console.WriteLine(GetComparableVersion("not a semver"));
      Console.WriteLine(GetComparableVersion("v1.2.3"));
      Console.WriteLine(GetComparableVersion("2.3.5-beta.2"));
      Console.WriteLine(GetComparableVersion("2.3.5"));
   }
   
   // official semver with start/end check removed
   // from https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
   private static readonly Regex semVerRegex = new Regex(@"(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?");

   internal static Version GetComparableVersion(string vString)
   {
       var match = semVerRegex.Match(vString);
       if (match.Success){
           // int.TryParse defaults to 0 if conversion fails, which is the desired behavior
           int.TryParse(match.Groups[1].Value, out var major);
           int.TryParse(match.Groups[2].Value, out var minor);
           int.TryParse(match.Groups[3].Value, out var build);
           return new Version(major, minor, build);
       }
       return new Version(0,0,0);
   }
}
// The example displays the following output:
//   0.0.0
//   1.2.3
//   2.3.5
//   2.3.5
```

src/NATS.Client/JetStream/ServerInfo.cs
```diff
@@ -65,5 +66,29 @@ public ServerInfo(string json)
...
+                int at = vString.IndexOf("-", StringComparison.Ordinal);
+                return "" + (Int32.Parse(v[0]) * 10000) + (Int32.Parse(v[1]) * 100) + Int32.Parse(v[2])
+                       + (at == -1 ? "" : vString.Substring(at));
+            }
+            catch (Exception) {
+                return "";
+            }
+        }
+
+        public bool IsNewerVersionThan(string vTarget) {
```
caleblloyd: I don't think these methods are necessary; could just make `GetComparableVersion` an `internal` function then use `<`, `>`, `==`, etc

scottf: Are you saying you can string do `if ("foo" > "bar") ...`

caleblloyd: No once `GetComparableVersion` returns a `Version` object then this would work:
```
if (serverInfo.GetComparableVersion() >= Version(2,7,3))
```

src/NATS.Client/JetStream/ServerInfo.cs
scottf: So I'm going to leave this as is. It does work already. I don't want to add regex. The code matches the java code so it's maintainable that way. The Version object doesn't have as friendly, self documenting functions as I already have. The Version class also is really only built to handle numbers, which is fine for released versions, but not for intermediate versions. Also, since we know this is a server version, it's safe to make assumptions.

caleblloyd: Ok...  I don't think the pre-release matching is working correctly currently though.  See my comment above.

src/NATS.Client/JetStream/ServerInfo.cs
caleblloyd: If these must be kept, they probably should not be `Public` so that we can change the method signature later if needed.  `Internal` should work, right?

scottf: I can't test them if I make them internal

src/Tests/UnitTests/JetStream/TestServerInfo.cs
scottf: Ok I fixed the code and the test. `2.3.5` comparable will be `20305~` which will be less than `20305-beta.2`

src/Tests/UnitTests/JetStream/TestServerInfo.cs
scottf: same as above

src/NATS.Client/JetStream/ServerInfo.cs
caleblloyd: The `.csproj` can make internals visible to other assemblies using [InternalsVisibleTo](https://docs.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-items?view=vs-2022#internalsvisibleto)

Example:

```
  <ItemGroup>
    <InternalsVisibleTo Include="$(AssemblyName).UnitTests" />
  <ItemGroup>
 ```

Note this is available starting from .NET SDK 5.0.  For older SDK versions, the [InternalsVisibleToAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=net-6.0) can be used

Comment (caleblloyd) [reply src/NATS.Client/JetStream/ServerInfo.cs]: The csproj already has the [InternalsVisibleToAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=net-6.0) for `UnitTests`

https://github.com/nats-io/nats.net/blob/59eca379ead07610244a385b4c110ce7987c13f3/src/NATS.Client/Properties/AssemblyInfo.cs#L7-L11

File: src/NATS.Client/KeyValue/KeyValue.cs [hunk 5]
```diff
@@ -172,25 +179,60 @@ public IList<string> Keys()
...
+                limit = DateTime.UtcNow.AddMilliseconds(-dmThresh);
+            }
+
+            IList<string> keepList0 = new List<string>();
+            IList<string> keepList1 = new List<string>();
+            VisitSubject(KeyValueUtil.ToStreamSubject(BucketName), DeliverPolicy.LastPerSubject, true, false, m =>
+            {
+                KeyValueEntry kve = new KeyValueEntry(m);
+                if (!kve.Operation.Equals(KeyValueOperation.Put)) {
+                    if (DateTime.Compare(kve.Created, limit) == 1) // created > limit, so created after
```
caleblloyd: should be able to just use `kve.Created > limit`, it calls `.Compare` under the covers and is easier to read

scottf: fixed

caleblloyd:
I think the only thing left is to make these 3 methods `internal` instead of `public`: `IsNewerVersionThan`, `IsSameVersion` and `IsOlderVersionThan`

caleblloyd (approved):
LGTM

PR #563, opened 2022-02-21
Issues 10, 91 StreamInfo request changes to retrieve subjects, deleted details

scottf:
[Issue 10: Only report full deleted list when requested](https://github.com/nats-io/nats-architecture-and-design/issues/10)
[Issue 91: StreamInfo request changes to retrieve subjects](https://github.com/nats-io/nats-architecture-and-design/issues/91)


src/NATS.Client/JetStream/IJetStreamManagement.cs
```diff
@@ -48,14 +48,21 @@ public interface IJetStreamManagement
...
         /// <returns>Stream information</returns>
         StreamInfo GetStreamInfo(string streamName);
 
+        /// <summary>
+        /// Get information about a stream, and include subject info.
+        /// </summary>
+        /// <param name="streamName">The name of the stream.</param>
+        /// <param name="subjectFilter">the subject filter to use. Can be wildcard. Null is equivalent to all subjects, the same as &gt;.</param>
+        /// <returns>Stream information</returns>
+        StreamInfo GetStreamInfoWithSubjectInfo(string streamName, string subjectFilter);
```
scottf: looking for suggestions for a better name for this. I didn't want to overload getStreamInfo b/c the behavior is different, namely the original does not get any subject info

ColinSullivan1: imo an overload makes sense w/ documentation and a slight change of behavior. This would be a good place for an optional parameter, but perhaps a future version - we'll want to maintain binary compatibility.

A suggestion: doc `GetStreamInfo(string streamName)` to indicate no subject info is returned and reference the overload if subject info is desired.  Then in this method (overload vs renamed) document that the subject filter will return subjects (use ">" for all).  imo the path of least surprise would be a null value for the filter would not return subject info just like the method above.

scottf: ok that's what I had originally but was not sure.

scottf: Ok so this is going to change differently now. In addition to subject info, there is also deleted details.

scottf:
Using administrative merge. PR has been available for 25 days. Parallel code has already been merged into the Java line.

PR #565, opened 2022-02-25
Minor fixes to spellings and output format of Options.ToString()

Wennemyr:


sixlettervariables (approved):
Good catch

scottf:
@Wennemyr Is it possible that you sign your commit? https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

src/NATS.Client/Options.cs
```diff
@@ -828,7 +828,7 @@ public override string ToString()
             }
             sb.AppendFormat("SubChannelLength={0};", SubChannelLength);
             sb.AppendFormat("Timeout={0};", Timeout);
-            sb.AppendFormat("Pendantic={0}", Pedantic);
+            sb.AppendFormat("Pedantic={0}", Pedantic);
```
scottf: Might as well remove this while we are here, it's just a duplication of line 801

scottf (changes requested):

src/NATS.Client/Options.cs
Wennemyr: I didn't know about signing, will figure it out before next time. My apologies, and thank you.

scottf (approved):

scottf:
using admin to merge unsigned commits for extremely minor text change.

PR #570, opened 2022-03-02
move from netstandard1.6 to netstandard2.0

### LINKED ISSUE #569: Move from netstandard1.6 to netstandard2.0
Moving from `netstandard1.6` to `netstandard2.0` would open a significant number of APIs including [ClientWebSocket](https://docs.microsoft.com/en-us/dotnet/api/system.net.websockets.clientwebsocket?view=netstandard-2.0)

https://docs.microsoft.com/en-us/dotnet/standard/net-standard

> .NET Standard 1.6 has 13,501 of the 37,118 available APIs.

> .NET Standard 2.0 has 32,638 of the 37,118 available APIs.

It would drop support for .NET Core 1.x, however maintenance ended on that [June 27, 2019](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core)

caleblloyd:
Resolves #569

### DIFF
--- a/src/NATS.Client/NATS.Client.csproj
+++ b/src/NATS.Client/NATS.Client.csproj
@@ -1,8 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks Condition="$(OS) == 'Windows_NT'">netstandard1.6;netstandard2.0;net46</TargetFrameworks>
-    <TargetFrameworks Condition="$(OS) != 'Windows_NT'">netstandard1.6;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks Condition="$(OS) == 'Windows_NT'">netstandard2.0;net46</TargetFrameworks>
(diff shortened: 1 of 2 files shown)


scottf:
@ColinSullivan1 @sixlettervariables 
Is this a backward compatible change or will it break existing users?

caleblloyd:
This removes support for .NET Core 1.0 and 1.1, both of which have been [End of Support since June 27, 2019](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core)

Compatibility chart is [here](https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0#select-net-standard-version)

JohannesEH:
I absolutely love this change. ⭐ 

I know this is extending the scope a bit and is an additional breaking change, but shouldn't the `net46` target be removed as well since official support for this version ends in a month (on Apr 26, 2022). 

Please note the `netstandard2.0` target supports .NET Framework 4.6.1+, so in my eyes it's probably not a big deal to remove 4.6 support. It might be nice to group these compatibility breaking changes together.

Also note that by removing the `net46` target we would be able to remove the "Windows_NT" conditions and remaining package references in the project file.

PR #579, opened 2022-03-09
msbuild: use TargetFrameworks exclusivley to fix overrides

caleblloyd:
Currently, mixing `<TargetFramework>` and `<TargetFrameworks>` means that when `Directory.Build.props` is merged, both keys can end up in the configuration, causing the following error on build on Linux:

```
Microsoft.PackageDependencyResolution.targets(267,5): error NETSDK1005: Assets file 'nats.net/src/NATS.Client/obj/project.assets.json' doesn't have a target for 'netcoreapp3.1'. Ensure that restore has run and that you have included 'netcoreapp3.1' in the TargetFrameworks for your project. [nats.net/src/NATS.Client/NATS.Client.csproj]
```

The solution is to only use `<TargetFrameworks>` so that overrides are always applied properly

### DIFF
--- a/src/Benchmarks/MicroBenchmarks/MicroBenchmarks.csproj
+++ b/src/Benchmarks/MicroBenchmarks/MicroBenchmarks.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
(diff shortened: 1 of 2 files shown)


caleblloyd:
This is pretty much the same change as #564  for a couple remaining instances of `<TargetFramework>`

tbeets:
LGTM

tbeets (approved):

PR #580, opened 2022-03-18
Js Multi Tool

scottf:
Ported to .NET. Also changed the Pull subscribe Fetch implementation based on behavior found while testing. The `JetStreamPullSubscription` class is the only library class that has changed. All other classes are non library projects.


src/Benchmarks/JsMultiConsumer/Consumer.cs
```diff
@@ -0,0 +1,51 @@
...
+                .BatchSize(100)
+                .MessageCount(100_000)
+                ;
+
+            a.PrintCommandLine();
+
+            Context ctx = new Context(a);
+
+            // Consumer sets up the stream for latency (non-normal) runs.
+            // Uncomment when
```
ColinSullivan1: ?

scottf: sorry checked this in unfinished. Am fixing

caleblloyd (approved):

PR #584, opened 2022-03-25
fixed kv key with dot as part of the key

scottf:
Same as [java Issue #621](https://github.com/nats-io/nats.java/issues/621)

### DIFF
--- a/src/NATS.Client/KeyValue/KeyValueUtil.cs
+++ b/src/NATS.Client/KeyValue/KeyValueUtil.cs
@@ -11,6 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using System.Text;
 using NATS.Client.Internals;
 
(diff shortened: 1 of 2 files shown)


src/NATS.Client/KeyValue/KeyValueUtil.cs
```diff
@@ -73,7 +74,14 @@ internal class BucketAndKey {
         public BucketAndKey(string subject) {
             string[] split = subject.Split('.');
```
caleblloyd: I think this could work as-is with `subject.Split(new []{'.'}, 3)` to limit the split array size to 3 elements

scottf: fixed

caleblloyd (approved):

kozlovic (approved):
LGTM

PR #587, opened 2022-04-01
tune kv subscribe supported functions keys / history / purge

scottf:
1. refactor added CalculatedPending since needed in more than 1 place
2. better design of visitSubject which is used for methods keys / history / purge

### DIFF
--- a/src/NATS.Client/KeyValue/KeyValue.cs
+++ b/src/NATS.Client/KeyValue/KeyValue.cs
@@ -255,24 +255,29 @@ namespace NATS.Client.KeyValue
             IJetStreamPushSyncSubscription sub = js.PushSubscribeSync(subject, pso);
             try
             {
-                int timeout = 5000; // give plenty of time for the first. Also used to quit loop
-                while (timeout > 0)
+                bool lastTimedOut = false;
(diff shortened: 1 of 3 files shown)


src/NATS.Client/KeyValue/KeyValue.cs
```diff
@@ -255,25 +255,30 @@ public KeyValueStatus Status()
...
+                        }
+                        lastTimedOut = false;
                     }
-                    else
+                    catch (NATSTimeoutException)
                     {
-                        timeout = 100; // the rest should come pretty quick
+                        if (lastTimedOut)
+                        {
+                            return; // two timeouts in a row is enough
```
caleblloyd, line 276: is it correct to swallow the exception?  or should this be `throw` instead of `return`?

scottf: I don't want the operations that use this to fail. They get the messages they get. I'm allowing for 1 timeout to account for the time between when the subscription is made and messages arrive at the client and never expect this condition to occur. 2 timeouts is arbitrary anyway. I'd like to push it as as, we could always change it later as the api contract does not prevent throwing an exception.

caleblloyd (approved):

PR #591, opened 2022-04-15
Allow setting CheckCertificateRevocation for encrypted connections

### LINKED ISSUE #492: Connect when CRL is missing in server certificate
Hi,

I am trying to use TLS when connecting to a local nats server. I am by no means an expert in TLS and just try to google my way to something working, so apologies if I have made some obvious mistake or misunderstood how it should be done. 

I have used the steps here https://docs.nats.io/nats-server/configuration/securing_nats/tls#creating-self-signed-certificates-for-testing to generate a certificate using mkcert and it is working fine using a browser.

Server is run with the following arguments

```bash
nats-server --tls --tlscert server.pem --tlskey server-key.pem --tlscacert rootCA.pem
```

And my test program is a simple connect and publish

```csh
using NATS.Client;

options.Secure = true;
using var connection = new ConnectionFactory().CreateConnection(options);
connection.Publish("Foo", Encoding.UTF8.GetBytes("Hello, World!"));
```

However, when trying to connect using this library I get a certificate validation error

```
RevocationStatusUnknown
An incomplete certificate revocation check occurred.
```

and it is due to the certificate not having a CRL configured, but this library enforces that connections should be checked for revocation.

https://github.com/nats-io/nats.net/blob/master/src/NATS.Client/Conn.cs#L446

```cs
SslProtocols protocol = (SslProtocols)Enum.Parse(typeof(SslProtocols), "Tls12");
sslStream.AuthenticateAsClientAsync(hostName, options.certificates, protocol, true).Wait();
```

I tested flipping the last flag to false, and connection works as expected but at the cost of not having CRL checking for any certificate instead. Ideally, the framework should offer something that only checks CRL if the server offers a certificate with CRL distribution points configured.

Obviously I can use `TLSRemoteCertificationValidationCallback` to bypass the error and everything works fine, but it would be non-trivial to write a correct handler for this specific situation.

So far not really an issue since it is just how .net works.

I tried writing the same program i go and to my surprise it connects just fine.

```go
package main

import (
	"log"
	"github.com/nats-io/nats.go"
)

func main() {

	nc, err := nats.Connect("tls://localhost:4222")
	if err != nil {
		log.Fatal(err)
	}
	defer nc.Close()

	nc.Publish("foo", []byte("Hello, World!"))
	nc.Flush()

	if err := nc.LastError(); err != nil {
		log.Fatal(err)
	}
}
```

My expectation here was that the two libraries would behave the same. 

Not a big issue but caused me to spend a lot of time troubleshooting why TLS was not working as expected so perhaps this can help others.

Perhaps a simple option to disable CRL checking could be added?

LaurensVergote:
This allows for use of self-signed certificates.
More info: https://en.wikipedia.org/wiki/Certificate_revocation_list

Gist with sample application using the new option:
https://gist.github.com/LaurensVergote/419aba52093b9cf7529d80cf4ae0af0f

This likely fixes #492

Adding @jens-gellynck (security analyst)

### DIFF
--- a/src/NATS.Client/Conn.cs
+++ b/src/NATS.Client/Conn.cs
@@ -443,7 +443,7 @@ namespace NATS.Client
                 try
                 {
                     SslProtocols protocol = (SslProtocols)Enum.Parse(typeof(SslProtocols), "Tls12");
-                    sslStream.AuthenticateAsClientAsync(hostName, options.certificates, protocol, true).Wait();
+                    sslStream.AuthenticateAsClientAsync(hostName, options.certificates, protocol, options.CheckCertificateRevocation).Wait();
                 }
                 catch (Exception ex)
                 {
--- a/src/NATS.Client/Options.cs
+++ b/src/NATS.Client/Options.cs
@@ -305,6 +305,8 @@ namespace NATS.Client
             {
                 certificates = new X509Certificate2Collection(o.certificates);
             }
+
+            CheckCertificateRevocation = o.CheckCertificateRevocation;
         }
 
         static readonly string[] protcolSep = new[] {"://"};
@@ -769,6 +771,12 @@ namespace NATS.Client
         /// <seealso cref="ReconnectJitter"/>
         /// <seealso cref="SetReconnectJitter(int, int)"/>
         public int ReconnectJitterTLS { get => reconnectJitterTLS; }
+
+        /// <summary>
+        /// Get or set whether to check Certificate Revocation when connecting via TLS
+        /// </summary>
+        /// <seealso cref="SslStream.AuthenticateAsClientAsync(string, X509CertificateCollection, System.Security.Authentication.SslProtocols, bool)"/>
+        public bool CheckCertificateRevocation { get; set; } = true;
 
         /// <summary>
         /// Returns a string representation of the


jens-gellynck (commented):
Looks good to me

sixlettervariables (approved):

scottf:
@LaurensVergote Does this cover OCSP Staples?

LaurensVergote:
> @LaurensVergote Does this cover OCSP Staples?

Not sure. Couldn't find anything specific about this on the [msdn page of this method](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslstream.authenticateasclientasync?view=net-6.0#system-net-security-sslstream-authenticateasclientasync(system-string-system-security-cryptography-x509certificates-x509certificatecollection-system-boolean))

Do you have an easy way of testing this? I worked together with @jens-gellynck on this so maybe he knows more?

jens-gellynck:
> @LaurensVergote Does this cover OCSP Staples?

The _checkCertificateRevocation_ parameter on _SslStream.AuthenticateAsClient_ will determine the _X509RevocationMode_ (checkCertificateRevocation ? X509RevocationMode.Online : X509RevocationMode.NoCheck) for the _X509Chain_. The documentation mentions a _CRL_ is used when **X509RevocationMode.Online** is used (https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509revocationmode?view=net-6.0). However, according to this stackoverflow post this will use OSCP when possible: https://stackoverflow.com/questions/56341594/does-x590certificate-build-use-ocsp-if-chainpolicy-revocationmode-online-is-used

Issue #389, opened 2020-06-12
Adding async versions of event handlers for connection establishment


ebekker:
The `Options` class allows a caller to register several event handlers (`EventHandler<T>`) to delegate various stages of authentication during connection establishment, via methods such as `SetUserCredentials(...)` or `SetNKey(...)`.

All of these take `EventHandler`s which means they are limited to synchronous code in the handler, but if the handler needs to call async code, there's no graceful way of doing this other than some sort of synchronization which would typically tie up a thread.

Perhaps it's time to look at adding asynchronous versions of these event handlers?

This would of course mean applying applying async pattern _all the way down_ starting with the `ConnectionFactory.Create*Connection(...)` family of functions, but that's pby a worthwhile change to modernize the library.

ebekker:
If this type of change would be welcome, then I would be happy to put together a PR?

ColinSullivan1:
@ebekker , Thanks for the offer!  Could you provide an small example with one of the event handlers so we can see what you are proposing without a significant time investment on your end?

ebekker:
Sure thing, let's say I want to customize my logic for handling the signing of the server nonce with NKeys authentication.  Using the `NATS.Client.Options` class, I would set a custom handler for this event:

```csharp
async Task ConnectAndPublish()
{
    var opts = ConnectionFactory.GetDefaultOptions();
    var jwt = await ResolveMyJWT(); // Get the JWT using some async approach
    opts.SetNKeys(jwt, HandleUserSigEvent);
    
    var cf = new ConnectionFactory();
    var conn = cf.CreateConnection(opts);
    conn.Publish("to.whom.it-may-concern", "Hello World!");
}

void HandleUserSigEvent(object source, UserSignatureEventArgs ev)
{
    // Because EventHandler<T> has a void return type, we have to do sync-over-async, ugh!
    var signed = SignChallengeNonce(ev.ServerNonce)
        .ConfigureAwait(false)
        .GetAwaiter()
        .GetResult();

    ev.SignedNonce = signed;
}

Task<byte[]> SignChallengeNonce(byte[] nonce)
{
    // TODO: does something async to retrieve NKeys and sign the nonce
}

ebekker:
Unfortunately, .NET BCL does not include an async version of [EventHandler](https://docs.microsoft.com/en-us/dotnet/api/system.eventhandler-1?view=netcore-3.1), but you could define your own:

```csharp
[...]

PR #592, opened 2022-04-19
[NO MERGE] Replace Monitor with SemaphoreSlim

jasper-d:
This is a first PR in a series of pull requests I have in mind to enable async IO in nats.net. This pull request in itself does not enable async, but is fundamental for it. It replaces all usages of `Monitor` (aka `lock`) in `Connection` with `SemaphoreSlim`.  While `Monitor` only works on sync paths, `SemaphoreSlim` enables sync and async waiting and can be used on async paths too. 

The diff is rather large because `SemaphoreSlim` is not reentrant, thus requiring some changes to when and where we lock on the semaphore. While refactoring, I used `Unsynchronized` as a method name suffix to indicate methods that require synchronization by the caller (precendence https://github.com/dotnet/runtime/search?l=C%23&q=Unsynchronized) and `Synchronized` for methods acquiring a lock themself (caller must not synchronize). Those suffixes can be removed, but I actually found them quite helpful.

I haven't tested these changes thoroughly yet, but unit tests are passing (some are flaky, I tried to fix some and added comments to others).
Plan is to have this PR sit here for some time, revisit it, and collect some feedback, as this is a rather risky change.

nats-io/nats.net#389 nats-io/nats.net.v2#48 nats-io/nats.net.v2#49

### DIFF
--- a/src/NATS.Client/AsyncSub.cs
+++ b/src/NATS.Client/AsyncSub.cs
@@ -12,6 +12,7 @@
 // limitations under the License.
 
 using System;
+using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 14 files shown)


jasper-d:
@ColinSullivan1 @scottf  Do you have an overview of what state changes need to be synchronized in `Connection`?

I reckon at least `status`, `bw`, `br`, `pending`, `waitingRequests`, `nextRequestId`, and `subs`. 

`Statistics` could be implemented thread safe itself, `ssid` should only be accessed using `Interlocked.*`. 

I'm not sure about `url`, `info`, and `lastEx`.

PR #594, opened 2022-04-25
changed to throw the second exception instead of hiding it

scottf:


scottf:
nope changed my mind again. I don't actually know if the timeout is an indication of a problem or an end of data

PR #597, opened 2022-05-04
subscription before consumer

scottf:
The order of creating a subscription on the server and creating a consumer on the server matters. Once the consumer is created, there is interest and the server tries to deliver. But if the subscription is not created, the messages are delivered to...nowhere, but are considered delivered.

This was not strictly a problem but it was a race - if the subscription was ready before the consumer was sent message, then things went fine. Unit test didn't fail. But when we were testing against NGS and in clusters with mixes of JetStream and non-Jetstream servers, the consumer was always ready because of simple latency.

So now the server subscription is always made first avoiding the problem altogether.


aricart (approved):
LGTM

PR #605, opened 2022-05-24
remove enforcement of durable name on pull subscriptions

scottf:
Since v2.7.0 the server allows ephemeral pull consumers. This change removes the code that enforced a durable name when making a pull subscription.


caleblloyd:
Could you add a description, linked issue, or link to the Client Spec change that explains why this change is being made?

PR #611, opened 2022-06-08
Ability to create StreamConfig objects Placement, External, Source, Mirror

scottf:
There was never a public way for a user to create StreamConfig objects Placement, External, Source, Mirror so you could only read the state from an existing stream.


tbeets (approved):
LGTM

PR #615, opened 2022-06-16
tls variations example

scottf:


scottf:
This is just a sample, no core code, merging without admin.

src/Samples/TlsVariationsExample/TlsVariationsExample.cs
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        private static Options RevocationCheckingOptions()
+        {
+            Options opts = ConnectionFactory.GetDefaultOptions();
+            opts.Url = Url;
+            opts.Secure = true;
+            X509Certificate2 cert = new X509Certificate2("client.pfx", "password");
+            opts.AddCertificate(cert);
+            opts.CheckCertificateRevocation = true;
```
LaurensVergote, line 58: true is the default so not really necessary
https://github.com/nats-io/nats.net/blob/master/src/NATS.Client/Options.cs#L791

PR #633, opened 2022-06-29
Accept encoded connection urls

scottf:


src/NATS.Client/Conn.cs
```diff
@@ -1331,21 +1331,15 @@ private string connectProto()
...
-                    string[] userpass = u.Split(':');
-                    if (userpass.Length > 0)
-                    {
-                        user = userpass[0];
-                    }
-                    if (userpass.Length > 1)
-                    {
-                        pass = userpass[1];
-                    }
+                    token = WebUtility.UrlDecode(u);
```
caleblloyd: I believe that `Uri.UnescapeDataString` is preferable to `WebUtility.UrlDecode`  because the former will not unescape a Plus sign to a Space

```
var uri = new Uri("http://test%3Auser+:test%3Apassword@localhost");
Console.WriteLine(Uri.UnescapeDataString(uri.UserInfo));
Console.WriteLine(WebUtility.UrlDecode(uri.UserInfo));
```

outputs:
```
test:user+:test:password
test:user :test:password
```

scottf: I was in the process of checking that and the unit test has both `%20` and `+` for space. The must have fixed it?

scottf: Apparently no one on the internet agrees, surprise there, but as example `https://user:pass@www.foo.com:8080/file;p=1?q=2#third` anything to the left of the query (left of the `?`) `+` means `+` and not a space, space must be encoded to `%20`. On the right side, `+` must be encoded otherwise it means space. Go figure. So user/pass is to the left, I'll switch to `Uri.UnescapeDataString` and fix the test

src/NATS.Client/Conn.cs
```diff
@@ -1410,6 +1404,9 @@ private string connectProto()
             if (opts.NoEcho && info.protocol < 1)
                 throw new NATSProtocolException("Echo is not supported by the server.");
 
+            var sb0 = new StringBuilder();
+            Console.WriteLine(info.AppendAsJsonTo(sb0));
```
caleblloyd: Debug code that needs to be removed?

src/NATS.Client/Conn.cs
caleblloyd: That sounds right...

sixlettervariables: I've had to support both to deal with terrible server implementations...

src/NATS.Client/Conn.cs
scottf: doh. fixed

caleblloyd (approved):
LGTM

caleblloyd:
Added a UTF-8 test case in https://github.com/nats-io/nats.net/pull/635/files that targets this branch, @scottf can I merge that into this branch and re-run CI?

Issue #629, opened 2022-06-27
Support consumer info fields changed to sequence_info from sequence_pair


scottf:
https://github.com/nats-io/nats-architecture-and-design/issues/122

PR #634, opened 2022-06-30
consumer info change to sequence_info from sequence_pair

scottf:
https://github.com/nats-io/nats.net/issues/629

* A new object was created, `SequenceInfo` which extends `SequencePair`
* The `ConsumerInfo.Delivered` and `ConsumerInfo.AckFloor` fields where changed to this type.

Backward compile compatibility was ensured by the subclass, meaning if you assigned `Delivered` to a SequencePair, this is fine, it still is a SequencePair.


RichardHightower (approved):
LGTM

PR #635, opened 2022-06-30
add utf8 test case to TestEncodedPassword

caleblloyd:
NOTE: Target branch is `accept-encoded-urls`

Add utf8 test case


caleblloyd:
update made in original PR

PR #644, opened 2022-07-20
Use Utc times for integration tests

LaurensVergote:
Noticed a few of the integration tests failed when ran locally and decided to fix them.
The issue was that local timestamps were being compared with utc timestamps.
In .NET the DateTimeKind is not checked when comparing, so we must make sure these are the same.

### DIFF
--- a/src/Tests/IntegrationTests/TestJetStreamManagement.cs
+++ b/src/Tests/IntegrationTests/TestJetStreamManagement.cs
@@ -31,7 +31,7 @@ namespace IntegrationTests
         {
             Context.RunInJsServer(c =>
             {
-                DateTime now = DateTime.UtcNow;
+                DateTime utcNow = DateTime.UtcNow;
 
                 IJetStreamManagement jsm = c.CreateJetStreamManagementContext();
 
@@ -42,7 +42,7 @@ namespace IntegrationTests
                     .Build();
 
                 StreamInfo si = jsm.AddStream(sc);
-                Assert.True(now <= si.Created);
+                Assert.True(utcNow <= si.Created.ToUniversalTime());
 
                 Assert.NotNull(si.Config);
                 sc = si.Config;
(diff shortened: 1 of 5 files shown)


sixlettervariables (approved):

src/Tests/IntegrationTests/TestJetStreamManagement.cs
```diff
@@ -31,7 +31,7 @@ public void TestStreamCreate()
         {
             Context.RunInJsServer(c =>
             {
-                DateTime now = DateTime.Now;
+                DateTime now = DateTime.UtcNow;
```
scottf: Now that you pointed this out I think it would be better to change the assertion on line 45 to normalize the data before comparison.

src/Tests/IntegrationTests/TestJetStreamManagement.cs
```diff
@@ -81,7 +81,7 @@ public void TestStreamCreate()
         public void TestStreamCreateWithNoSubject() {
             Context.RunInJsServer(c =>
             {
-                DateTime now = DateTime.Now;
+                DateTime now = DateTime.UtcNow;
```
scottf: Same comment on comparison line 94

src/Tests/IntegrationTests/TestJetStreamManagement.cs
```diff
@@ -639,7 +639,7 @@ public void TestGetStreams()
                 MsgHeader h = new MsgHeader();
                 h.Add("foo", "bar");
 
-                DateTime beforeCreated = DateTime.Now;
+                DateTime beforeCreated = DateTime.UtcNow; //MessageInfo.Time is in UTC
```
scottf, line 642: fix the comparison line 652, 660

scottf (dismissed):

scottf:
@LaurensVergote Thanks for this. Can you please sign your commits.  https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

src/Tests/IntegrationTests/TestJetStreamManagement.cs
LaurensVergote: As a rule of thumb, it is best to work in UTC everywhere in the code until the point it needs to be displayed/logged. This way you avoid any confusion and potential errors due to winter/summer times.
I'll see what I can do for this commit

arondi (approved):
LGTM

scottf:
Admin merge - signing was added on most recent commits.

Issue #645, opened 2022-07-23
Consumer configuration num_replicas and mem_storage


scottf:
https://github.com/nats-io/nats-architecture-and-design/issues/130
https://github.com/nats-io/nats-architecture-and-design/issues/134

PR #647, opened 2022-07-25
Consumer replicas mem storage

scottf:
https://github.com/nats-io/nats.net/issues/645

support num_replicas and mem_storage in consumer configuration


RichardHightower (approved):
LGTM

PR #653, opened 2022-08-04
no_erase

scottf:
https://github.com/nats-io/nats-architecture-and-design/issues/118


src/NATS.Client/JetStream/JsonSerializable.cs
```diff
@@ -7,7 +7,7 @@ public abstract class JsonSerializable
     {
         internal abstract JSONNode ToJsonNode();
 
-        internal byte[] Serialize()
+        public byte[] Serialize()
```
arondi, line 10: Why was this done?

arondi (approved):
LGTM

src/NATS.Client/JetStream/JsonSerializable.cs
scottf: 1. It made it easier to test.
2. It a read of the data so it's okay to expose as public.

PR #654, opened 2022-08-05
Adjustments to interfaces to improve the mockability and faking of NATS for unit testing in consumer apps.

sspates:
Updated a few interfaces to use IConnection instead of Connection as Connection is internal.  This is preventing mocking libraries from being able to mock properly.
Added interface for ConnectionFactory so factory can be mocked/faked in unit tests.


src/NATS.Client/ISubscription.cs
```diff
@@ -70,7 +70,7 @@ public interface ISubscription : IDisposable
         /// <summary>
         /// Gets the <see cref="NATS.Client.Connection"/> associated with this instance.
         /// </summary>
-        Connection Connection { get; }
+        IConnection Connection { get; }
```
scottf: This would not be backward compatible.

src/NATS.Client/Subscription.cs
```diff
@@ -137,7 +137,7 @@ internal void close(bool closeChannel)
         /// <summary>
         /// Gets the <see cref="Connection"/> associated with this instance.
         /// </summary>
-        public Connection Connection => conn;
+        public IConnection Connection => conn;
```
scottf: Also not backward compatible.

scottf (changes requested):
I like this idea but we cannot change the interface in a non backward compatible way.

sspates:
@scottf I removed the IConnection changes, but found that protecting the Message in the handler makes mocking the handling of the message impossible.  I adjusted the Handler to expose the message to being set.

src/NATS.Client/IConnectionFactory.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// <param name="opts">The NATS client options to use for this connection.</param>
+        /// <returns>An <see cref="IEncodedConnection"/> object connected to the NATS server.</returns>
+        /// <exception cref="NATSNoServersException">No connection to a NATS Server could be established.</exception>
+        /// <exception cref="NATSConnectionException"><para>A timeout occurred connecting to a NATS Server.</para>
+        /// <para>-or-</para>
+        /// <para>An exception was encountered while connecting to a NATS Server. See <see cref="Exception.InnerException"/> for more
+        /// details.</para></exception>
+        IEncodedConnection CreateEncodedConnection(Options opts);
+    }
+}
```
scottf: nit: newline at end of file

src/NATS.Client/IConnectionFactory.cs
```diff
@@ -0,0 +1,143 @@
+﻿using System;
```
scottf: needs license at start of file

src/NATS.Client/Msg.cs
```diff
@@ -233,7 +233,7 @@ public void Respond(byte[] data)
                 throw new NATSException("No Reply subject");
             }
 
-            Connection conn = ArrivalSubscription?.Connection;
+            IConnection conn = ArrivalSubscription?.Connection;
```
scottf: Why does this change?

src/NATS.Client/NATS.cs
```diff
@@ -496,15 +496,10 @@ internal class IC
...
-        internal Msg msg = null;
-
         /// <summary>
         /// Retrieves the message.
         /// </summary>
-        public Msg Message
-        {
-            get { return msg; }
-        }
+        public Msg Message { get; set; }
```
sixlettervariables: Is there a reason for the setter to be public?

scottf:
@sspates-starbucks I'm good with the event message changes, not excited about having a public setter, but I suppose it's benign, the worst that could happen is the dev overwrites their own message.
Frankly I'm not seeing the need for the IConnectionFactory interface. There is already an IConnection interface so you can provide your own mock connection if you want to.
And I'll mention again the ability to run a server locally during tests. In .net integration tests we generally start a server for each suite, but for java we start a server for each test.

scottf (approved):
LGTM

src/NATS.Client/NATS.cs
sspates: Fixed

PR #659, opened 2022-08-13
Internal, automatic use of get direct

scottf:


src/NATS.Client/JetStream/JetStreamManagement.cs
```diff
@@ -150,31 +150,55 @@ public IList<StreamInfo> GetStreams()
...
-            return RequestResponseRequired(subj, messageGetRequest.Serialize(), Timeout);
+            CachedStreamInfo csi = GetCachedStreamInfo(streamName);
+            if (csi.AllowDirect) {
+                string subj;
+                byte[] payload;
+                if (messageGetRequest.IsLastBySubject) {
+                    subj = string.Format(JetStreamConstants.JsapiDirectGetLast, streamName, messageGetRequest.LastBySubject);
+                    payload = null;
+                }
+                else{
```
sixlettervariables:
```suggestion
                else {
```

sixlettervariables (approved):
Just a nit.

src/NATS.Client/JetStream/JetStreamManagement.cs
scottf: fixed

PR #660, opened 2022-08-16
general testing improvements

scottf:
it's flapping, not sure why the comparison is failing even though things are normalized to utc.


arondi (approved):
LGTM

PR #670, opened 2022-09-15
consumer create for server v2.9.0

scottf:


arondi (approved):
LGTM. Looks like a port of the java, including unit tests.

PR #677, opened 2022-09-19
object store tuning

scottf:
* specific error for link not allowed in meta used for putting and object
* object store adr changes to getInfo


ColinSullivan1 (approved):
LGTM.

PR #708, opened 2022-11-08
stream state lost

scottf:
https://github.com/nats-io/nats.net/issues/682 support stream state -> lost stream data

### DIFF
--- a/src/Tests/UnitTests/Data/StreamInfo.json
+++ b/src/Tests/UnitTests/Data/StreamInfo.json
@@ -32,13 +32,21 @@
     "first_seq": 13,
     "first_ts": "0001-01-01T00:00:00Z",
     "last_seq": 14,
-    "last_ts": "2021-01-01T00:00:00Z",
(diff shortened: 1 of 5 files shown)


src/Tests/UnitTests/Data/StreamInfo.json
```diff
@@ -32,13 +32,21 @@
...
     "consumer_count": 15,
     "num_subjects": 3,
+    "num_deleted": 6,
     "subjects": {
       "sub0": 1,
       "sub1": 2,
       "x.foo": 3
+    },
+    "deleted": [91, 92, 93, 94, 95, 96],
+    "lost": {
```
sixlettervariables, line 45: Do we have a test stream info without this property? Tried verifying on my phone but wasn't able to easily look around. Otherwise looks good.

scottf: I could probably figure it out. I keep this json up to date whenever I add a property, it gives me coverage

PR #724, opened 2022-12-21
validate version semver

scottf:


src/NATS.Client/Internals/Validator.cs
```diff
@@ -579,5 +580,22 @@ public static Duration EnsureDurationNotLessThanMin(long providedMillis, Duratio
         public static string EnsureEndsWithDot(string s) {
             return s == null || s.EndsWith(".") ? s : s + ".";
         }
+        
+        const string SemVerPattern = "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$";
```
aricart, line 584: this is expecting that they gave you some version that is exactly what you want (which is possibly ok), but you may also consider that if they give you "v1.2.3-test", that is valid as well once you remove the "v" and the qualifier part ("-test"). `(\d+).(\d+).(\d+)`

aricart (approved):
LGTM - save the nit, that it can possibly be simpler.

Issue #727, opened 2022-12-28
Add Async versions for all JetStream/KeyValue/ObjectStore APIs


iskandersierra:
## Feature Request

Add Async versions for all JetStream/KeyValue/ObjectStore APIs

#### Proposed Change:

- Add `async` variant for method `JetStreamBase.RequestResponseRequired` using `IConnection.RequestAsync` method
- Based on that new variant, add new `async` variants for methods on `JetStreamManagement`
- Add `async` variant for method `KeyValue._write` using `JetStream.PublishAsync`
- Based on that new variant, add new `async` variants for methods `Get`, `Put`, `Create`, `Update`, ... on `KeyValue` and `KeyValueManagement`
- Use the same approach to update `ObjectStore`

#### Who Benefits From The Change(s)?

Every developer using an `async` framework and the standard .net Task Pool, where they will avoid locking pooled threads for IO operations involving NATS, enhancing general throughput.

#### Alternative Approaches

Haven't thought of an alternative that does not lock pooled threads.

I propose myself for creating the PR for this issue, in case this effort cannot be developed by the current contributors.

scottf:
@iskandersierra We welcome PRs and submissions. I can't guarantee acceptance, but this would just be adding functionality, so we'll see. One thing to consider is that this client is based on older .NET and is not really fully async all the way through to the socket layer. I'm definitely interested in what you might come up with.

Maybe consider making a IKeyValueAsync interface as a parallel to IKeyValue?

Thanks!

iskandersierra:
Hi @scottf . I'll try my best to dig a bit around. For now I just want to familiarize myself with the internals, and at least add the `async` variants of the current APIs. I think that having `IKeyValueAsync` and `IObjectStoreAsync` is better than duplicating the API surface of the existing classes, so I'll go with that idea.

For the moment I will not touch the socket access layer. I'm just adding new APIs to make the less amount of noise

I'll publish a draft PR as soon as I have any noticeable advances.

iskandersierra:
I published the draft here #730

scottf:
Closed as will not be completed. This can be done manually. Also try https://github.com/nats-io/nats.net.v2

PR #730, opened 2022-12-29
Add JetStream async APIs

iskandersierra:
I want to start by reviewing the changes (style, approach, tests) before continuing. 

@scottf Let me know of any suggestions on the changes so far


UliKu-philips:
Hello, this is an interesting contribution. I have some feedback:
- [Consider eliding `async` and `await` when just passing through a `Task`](https://blog.stephencleary.com/2016/12/eliding-async-await.html)
- [Add `ConfigureAwait(false)` to awaited Tasks](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2007)

iskandersierra:
Hi @UliKu-philips 

>     * [Consider eliding `async` and `await` when just passing through a `Task`](https://blog.stephencleary.com/2016/12/eliding-async-await.html)

This change makes a trade-off between readability and performance. Using `async`/`await` when it can be pass-through, is less performant, but it properly captures the exception stack trace in case it happens. So only in cases where there is a proven performance issue and it is the more preferable attribute to ask, then I use pass-through Task.

>     * [Add `ConfigureAwait(false)` to awaited Tasks](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2007)

I oversaw this, so I'll change it in a next iteration

src/Benchmarks/JsMulti/JsMultiTool.cs
```diff
@@ -179,7 +179,7 @@ private static void _pub(Context ctx, Stats stats, int id, Publisher<PublishAck>
             }
             else
             {
-                publisher = js.PublishAsync;
+                publisher = (s, p) => js.PublishAsync(s, p);
```
scottf, line 182: Was this done because of an api change? If so, it's not acceptable. All changes must be completely backward compatible and require no change to user code.

src/NATS.Client/JetStream/IJetStreamManagementAsync.cs
```diff
@@ -0,0 +1,202 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NATS.Client.JetStream
+{
+    /// <summary>
+    /// This is the JetStream management async API to programmatically create, delete,
+    /// and update various JetStream objects.
+    /// </summary>
+    public interface IJetStreamManagementAsync
```
scottf, line 24: Can you justify the need for async management tasks? I don't really see this as necessary. I'll ask my team for other opinions.

src/NATS.Client/KeyValue/IKeyValueAsync.cs
```diff
@@ -0,0 +1,147 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NATS.Client.KeyValue
+{
+    public interface IKeyValueAsync
```
scottf, line 20: Maybe these can just get put in the normal interface instead of a whole new interface. Also, I'm not completely sure of the need for these.

src/NATS.Client/JetStream/IJetStream.cs
```diff
@@ -168,7 +169,7 @@ public interface IJetStream
         /// <param name="data">An array of type <see cref="byte"/> that contains the data to publish
         /// to the connected NATS server.</param>
         /// <returns>PublishAck</returns>
-        Task<PublishAck> PublishAsync(string subject, MsgHeader headers, byte[] data);
+        Task<PublishAck> PublishAsync(string subject, MsgHeader headers, byte[] data, CancellationToken cancellationToken = default);
```
scottf, line 172: I wouldn't worry about it yet, but will need to update the document comments if this is going to be accepted.

src/NATS.Client/JetStream/IJetStreamManagementAsync.cs
iskandersierra: In dotnet, when you do not use async code inside an async workflow, you are allowing threads in the thread pool (limited resource) to block for IO operations. In such cases you are reducing the throughput of the service by reducing the amount of unblocked threads in the shared pool.

By using `await client.OperationAsync(..., cancellationToken)` you are allowing the thread to return to the pool to serve new incoming requests meanwhile the `OperationAsync` have some result/error/cancellation to signal.

The async/non-async duality in dotnet is due to the backward compatibility nature of the dotnet framework. In fact, I'm of the opinion that any new library doing IO should only offer async operations, and leave to the client code the responsibility of blocking to their own peril.

With this PR I'm trying to introduce mirror async operations to the old sync-ones, because this library was first developed a long way ago and was not designed with this mindset. I would even suggest to redesign the library from the ground up, but it is not mi intention to disrupt existing maintainers and users.

src/NATS.Client/KeyValue/IKeyValueAsync.cs
iskandersierra: As per your suggestion [here](https://github.com/nats-io/nats.net/issues/727#issuecomment-1366957115) I did it this way

src/NATS.Client/JetStream/IJetStream.cs
iskandersierra: This is a change in public API. I believe all public Async APIs should have a `CancellationToken` as last optional argument, but it could be a breaking change for some rare cases, like using the method as a delegate instead of calling it with parameters.

src/NATS.Client/JetStream/IJetStreamManagementAsync.cs
ColinSullivan1: @iskandersierra, a redesign has been discussed. This version of the .NET client is admittedly old - it was originally built to compile on .NET 3.5, and was available even before .NET core 1.0 was released!  Right or wrong, this speaks to the dedication the .NET frameworks have toward backward compatibility.  We strive to do the same, so we do not change existing public APIs within a major release version (1.x.x).

We do understand that in order to modernize this a rewrite is necessary for exactly the reason you mentioned -  async should go all the way down to the I/O. There are also conventions that have changed and optimizations that we can take advantage of.  Because most of these end up blocking at I/O anyhow, I'd suggest encapsulation with an async interface until NATS.NET v2.0 is available.

Also, your input on what you'd like to see in a modern client is valuable to us so please let us know what you'd like to see!

src/NATS.Client/JetStream/IJetStream.cs
scottf: So it shouldn't change the api if there is a default for new values. But it did appear to affect the multi tool

iskandersierra: Ok, given that this changes are going to be included into the old-style API, I'll try to keep it as close to it as possible. Or maybe it is not very productive to include this changes in current library.

Is there any discussion place to talk about a new API definition for v2.0, as mentioned by @ColinSullivan1 above? I would also like to work on an implementation for F#.

src/NATS.Client/JetStream/IJetStreamManagementAsync.cs
iskandersierra: Perfect, I mention you in comments below. I would like to collaborate on API design and implementation as much as possible, also on a F# library, preferably native.

iskandersierra: Maybe there is no point on continuing this PR because no amount of async APIs will achieve real non-blocking calls to IO. I prefer to redirect efforts on developing a modern version for NATS.NET

src/NATS.Client/JetStream/IJetStream.cs
ColinSullivan1: I just created a discussion:  https://github.com/nats-io/nats.net/discussions/741

PR #734, opened 2023-01-23
Service multi endpoints

scottf:


src/NATS.Client/JetStream/JsonSerializable.cs
```diff
@@ -1,18 +1,31 @@
...
     {
-        internal abstract JSONNode ToJsonNode();
+        public abstract JSONNode ToJsonNode();
 
         public virtual byte[] Serialize()
         {
             return JsonUtils.Serialize(ToJsonNode());
         }
 
-        public override string ToString()
```
scottf, line 15: This was not released even though it was pushed to main.

src/NATS.Client/Msg.cs
```diff
@@ -108,6 +108,18 @@ public Msg(string subject)
...
+            subject = msg.subject;
+            sid = msg.sid;
+            _reply = msg._reply;
+            data = msg.data;
+            sub = msg.sub;
+            header = msg.header;
+            status = msg.status;
+            _lastAck = msg._lastAck;
+        }
+
```
scottf, line 122: Constructor needed to subclass

PR #742, opened 2023-02-12
Full Server List Provider

scottf:


src/NATS.Client/NatsServerListProvider.cs
```diff
@@ -0,0 +1,112 @@
...
+                }
+                if (removed) {
+                    _list.Add(lastConnectedServer);
+                }
+            }
+            return new List<NatsUri>(_list); // return a copy
+        }
+
+        private bool Add(NatsUri nuri) {
+            // TODO ResolveHostnames System.Net.Dns not available, ask for help
```
caleblloyd, line 76: I think leaving DNS Resolution in `Connect` is fine, and it shouldn't be done here.

If I understand correctly, this is a long-lived List and the TTL on DNS Entries could expire by the time the Client goes to connect to one of the alternate URLs.  Resolving in `Connect` like we do now means that it will get re-resolved each time and any DNS changes will get picked up.  Converting them to IP Addresses here will cause them to never be re-resolved during Connect.

caleblloyd: I see that it is guarded by `options.ResolveHostnames` though...  So if you want to do it here for `netstandard1.6` I believe the only way is with an async call (would need to be async-over-sync in this situation):

```
var addresses = Dns.GetHostAddressesAsync(hostName).GetAwaiter().GetResult();
```

scottf: the list is resolved once for the connect and every time for reconnect. I could move resolve out of the get list and instead resolve just before I try it, which means I might have 1 or more to try, but then I also have to add to the hook so the user can resolve for me. I will try that and also try the GetHostAddressesAsync

PR #764, opened 2023-05-04
Fixed TcpClient not disposed when an exception is thrown during connect

BrentSLC:
When an exception occurred during the ConnectAsync (i.e. socketException), the TcpClient would not be closed. In certain situations where the reconnect keeps failing with an exception this would result in an increase in handles as the ~TcpClient() does not clean up managed resources such as sockets.


scottf (approved):
LGTM

PR #767, opened 2023-05-10
Enable custom TCPConnection implementation

tkeller-moxe:
This PR expands on the TCPConnection class to be an interface that can be overridden with a options argument to supply a custom interface for implementing the connection class used by the nats client. 

The changes here are modeled after how the Java client handles a custom dialer. 

I included an example of my use case here which is enabling a TLS Reverse proxy to terminate the TLS connection by performing the SSL upgrade before nats communication starts.

### DIFF
--- a/src/NATS.Client/Connection.cs
+++ b/src/NATS.Client/Connection.cs
@@ -183,7 +183,7 @@ namespace NATS.Client
         // 60 second default flush timeout
         static readonly int DEFAULT_FLUSH_TIMEOUT = 10000;
 
-        TCPConnection conn = new TCPConnection();
+        ITCPConnection conn = new TCPConnection();
 
         SubChannelPool subChannelPool = null;
 
@@ -336,7 +336,8 @@ namespace NATS.Client
         /// Convenience class representing the TCP connection to prevent 
         /// managing two variables throughout the NATs client code.
         /// </summary>
-        private sealed class TCPConnection : IDisposable
+
+        public class TCPConnection : ITCPConnection
         {
             /// A note on the use of streams.  .NET provides a BufferedStream
             /// that can sit on top of an IO stream, in this case the network
@@ -354,15 +355,19 @@ namespace NATS.Client
             ///          ->NetworkStream/SslStream (srvStream)
             ///              ->TCPClient (srvClient);
             /// 
-            object        mu        = new object();
-            TcpClient     client    = null;
-            NetworkStream stream    = null;
-            SslStream     sslStream = null;
+            Options options;
 
-            string        hostName  = null;
+            object mu = new object();
+            TcpClient client = null;
+            NetworkStream stream = null;
+            SslStream sslStream = null;
 
-            internal void open(Srv s, int timeoutMillis)
+            string hostName = null;
+
+            public virtual void open(Srv s, Options options)
             {
+                this.options = options;
+
                 lock (mu)
                 {
                     // If a connection was lost during a reconnect we 
@@ -384,12 +389,12 @@ namespace NATS.Client
 
                     var task = client.ConnectAsync(s.Url.Host, s.Url.Port);
                     // avoid raising TaskScheduler.UnobservedTaskException if the timeout occurs first
-                    task.ContinueWith(t => 
+                    task.ContinueWith(t =>
                     {
                         GC.KeepAlive(t.Exception);
                         close(client);
                     }, TaskContinuationOptions.OnlyOnFaulted);
-                    if (!task.Wait(TimeSpan.FromMilliseconds(timeoutMillis)))
+                    if (!task.Wait(TimeSpan.FromMilliseconds(options.Timeout)))
                     {
                         close(client);
                         client = null;
@@ -398,9 +403,9 @@ namespace NATS.Client
 
                     client.NoDelay = false;
 
-                    client.ReceiveBufferSize = defaultBufSize*2;
-                    client.SendBufferSize    = defaultBufSize;
-                    
+                    client.ReceiveBufferSize = defaultBufSize * 2;
+                    client.SendBufferSize = defaultBufSize;
+
                     stream = client.GetStream();
 
                     // save off the hostname
@@ -420,18 +425,20 @@ namespace NATS.Client
                 return false;
             }
 
-            internal static void close(TcpClient c)
+            public virtual void close(TcpClient c)
             {
 #if NET46
                     c?.Close();
 #else
-                    c?.Dispose();
+                c?.Dispose();
 #endif
                 c = null;
             }
 
-            internal void makeTLS(Options options)
+            public void makeTLS()
             {
+                
+
                 RemoteCertificateValidationCallback cb = null;
 
                 if (stream == null)
@@ -460,7 +467,7 @@ namespace NATS.Client
                 }
             }
 
-            internal int SendTimeout
+            public int SendTimeout
             {
                 set
                 {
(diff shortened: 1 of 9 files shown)


tkeller-moxe:
Rebased, bringing back in line with main.

scottf:
@tkeller-moxe can you rebase or merge. Should fix the failing tests

src/Samples/TLSReverseProxyExample/TLSReverseProxyExample.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <RootNamespace>NATSExamples</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\NATS.Client\NATS.Client.csproj" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <NoWarn>$(NoWarn);NU1701</NoWarn>
+  </PropertyGroup>
```
scottf, line 17: What is this?

src/NATS.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29230.47
+# Visual Studio Version 17
```
scottf: Is this necessary to change this file version?

src/NATS.Client/NATS.Client.csproj
```diff
@@ -14,6 +14,7 @@
     <PackageTags>CNCF NATS Messaging Cloud Publish Subscribe PubSub</PackageTags>
     <DebugType>pdbonly</DebugType>
     <DebugSymbols>true</DebugSymbols>
+    <TargetFrameworks>netstandard1.6;netstandard2.0;net46</TargetFrameworks>
```
scottf: Please explain this.

src/Samples/TLSReverseProxyExample/CustomTCPConnection.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace NATSExamples
+{
+
+
+        /// <summary>
+        /// Convenience class representing the TCP connection to prevent 
+        /// managing two variables throughout the NATs client code.
+        /// 
+        /// This "Custom" implementation just makes the connection TLS after opening it. 
+        /// </summary>
```
scottf, line 22: Versus waiting for the normal progression of things? Is this why this was done in the first place?

src/NATS.sln
tkeller-moxe: I can revert this, I just have a newer version of Visual Studio.

src/NATS.Client/NATS.Client.csproj
tkeller-moxe: I think this was also VS, I'll remove.

src/Samples/TLSReverseProxyExample/CustomTCPConnection.cs
tkeller-moxe: Yes. My use case is to have a way to start the TLS connection BEFORE the traditional NATS hand shake allowing the connection to go through a load balancer. This PR ultimately tries to mimic the concept of the Custom Dialer used in the Java/Go implementations.

src/Samples/TLSReverseProxyExample/TLSReverseProxyExample.csproj
tkeller-moxe: This example was written in .netcore6, this has it drop the warnings saying that "Package with Version NET 4.6 Maynot be compatible." I can remove this if its a concern as well :) this was another instance of VS2022 being..."helpful"

tkeller-moxe:
@scottf just following back up! Let me know what other clarity you need.

scottf:
This is temporarily on hold while our team discusses it. There is already an initiative to change the server behavior and we are considering that this behavior might become the default behavior instead of an add on, but there are consideration such as handling old server versions when the change is made. I know that doesn't help your timeline and I'm not forgetting about it.

tkeller-moxe:
Understood, Let me know if there is anything I can help with on my end. 

Also an additional comment/musing, I have not tested it nor do I have a strong use case for it. This should actually enable any interface that can mimic the stream class to be used for communication. EG a unix socket is appropriate. Not tested/might want to get some renaming done if that's the remaining use case but I think this is generic enough to cover that as well.

tkeller-moxe:
I would like to bump this again, It is coming back up in security and networking audits on my end again. (And I would like to discuss before fixing the merges :) )
Thanks!

src/Samples/TLSReverseProxyExample/CustomTCPConnection.cs
scottf: @tkeller-moxe Do you mean custom SocketDataPort implementation?

tkeller-moxe: I think that is another proper name for it? 
https://docs.nats.io/using-nats/developer/tutorials/custom_dialer This is what its mimicing.

mtmk:
Difficult to see the actual change from the diff because of the class extract. For clarity, can we not do the refactor at this stage? (Seems to be causing conflicts as well.)

src/NATS.Client/Conn.cs
```diff
@@ -332,262 +332,6 @@ internal Control(string s)
             }
         }
 
-        /// <summary>
-        /// Convenience class representing the TCP connection to prevent 
-        /// managing two variables throughout the NATs client code.
-        /// </summary>
-        private sealed class TCPConnection : IDisposable
```
mtmk: Can we keep this class here? IMHO it's fine to stay as a nested class if necessary. I just want to make sure I can see all the changes clearly since mistakes here could create massive bugs.

mtmk (changes requested):
Please revert TCPConnection class extract refactor for clarity.

src/NATS.Client/Conn.cs
tkeller-moxe: I'll see what I can do, I moved it out for code clarity. 
If I remember I needed to unseal this class and add an interface and IMHO at that point I felt Conn had to many classes and responsibilities. Should be doable I'll try to have this in today :)

mtmk:
Thanks for the change @tkeller-moxe

Just to note, I see you needed to remove `sealed` from the class so you can extend it. That might introduce overhead (call vs. callvirt, new optimizations etc.) but I doubt it'd be noticeable and probably don't even apply most of the time.

LGTM but would be nice to resolve the conflicts. Are they straight forward to resolve?

src/NATS.Client/Connection.cs
```diff
@@ -420,22 +425,24 @@ internal void open(Srv s, int timeoutMillis)
...
-            internal void makeTLS(Options options)
+            public void makeTLS()
             {
+                Options options = natsConnection.Opts;
+
                 RemoteCertificateValidationCallback cb = null;
 
                 if (stream == null)
-                    throw new NATSException("Internal error:  Cannot create SslStream from null stream.");
+                    throw new NATSException("public error:  Cannot create SslStream from null stream.");
```
mtmk: This should stay as 'Internal error'

src/NATS.Client/Connection.cs
```diff
@@ -354,15 +355,19 @@ private sealed class TCPConnection : IDisposable
...
+            object mu = new object();
+            TcpClient client = null;
+            NetworkStream stream = null;
+            SslStream sslStream = null;
 
-            string        hostName  = null;
+            string hostName = null;
 
-            internal void open(Srv s, int timeoutMillis)
+            public virtual void open(Srv s, Connection connection, int timeoutMillis)
```
mtmk: I suggest to change this signature to:
```
void open(Srv s, Options opts)
```
`Connection` is only used to access options and timeout is part of options already.

mtmk (changes requested):
Looks good. I have a couple of small requests if that's ok. Should be a quick commit hopefully. Because it's becoming a public interface I thought a little extra care might be good.

mtmk (approved):
LGTM

PR #772, opened 2023-05-15
Fixed the message for status which says gets or sets when in fact we …

m0hammadb:
the message says gets or sets. if the intended behavior was also to set the status we have to add the setter otherwise we remove the set from the helper message which is confusing


scottf:
@m0hammadb thanks!

PR #781, opened 2023-06-07
Remove client side check against server info max_payload when publishing

scottf:


src/NATS.Client/Connection.cs
```diff
@@ -99,7 +99,7 @@ internal enum ClientProtcolVersion
     // encoded conn rather than using this class as
     // a base class.  This can happen anytime as we are using
     // interfaces.
-    public class Connection : IConnection, IDisposable
```
scottf, line 102: IConnection is already IDisposable

src/NATS.Client/Connection.cs
```diff
@@ -99,7 +99,7 @@ internal enum ClientProtcolVersion
...
+    public class Connection : IConnection
```
scottf, line 102: While I was in here, IConnection is already IDisposable

src/NATS.Client/Connection.cs
```diff
@@ -2597,10 +2597,6 @@ internal void PublishImpl(string subject, string reply, MsgHeader inHeaders, byt
                 if (isDrainingPubs())
                     throw new NATSConnectionDrainingException();
 
-                // Proactively reject payloads over the threshold set by server.
-                if (count > info.MaxPayload)
-                    throw new NATSMaxPayloadException();
-
```
scottf, line 2603: This is the change. `info` is Server info. Simply don't check it.

src/NATS.Client/Connection.cs
```diff
@@ -2680,8 +2676,6 @@ internal void PublishImpl(string subject, string reply, MsgHeader inHeaders, byt
         /// to the connected NATS server.</param>
         /// <exception cref="NATSBadSubscriptionException"><paramref name="subject"/> is 
         /// <c>null</c> or entirely whitespace.</exception>
-        /// <exception cref="NATSMaxPayloadException"><paramref name="data"/> exceeds the maximum payload size
-        /// supported by the NATS server.</exception>
```
scottf, line 2684: Removed api doc since this exception is not thrown anymore. There are 34 total of this exact change.

src/NATS.Client/Exceptions.cs
```diff
@@ -197,8 +198,8 @@ public class NATSNoRespondersException : NATSTimeoutException
     /// </summary>
     public class NATSJetStreamStatusException : NATSException
     {
-        private Subscription Sub { get; }
-        private MsgStatus Status { get; }
+        public Subscription Sub { get; }
+        public MsgStatus Status { get; }
```
scottf, line 202: Should have always been public.

src/Tests/IntegrationTests/TestBasic.cs
```diff
@@ -2082,7 +2082,31 @@ public void TestMessageHeaderNoServerSupport()
...
+                        {
+                            size = maxPayload + x;
+                            c.Publish("mp", new byte[size]);
+                            Thread.Sleep(100);
+                        }
+                    }
+                }
+            });
+        }
+
```
scottf, line 2109: The only thing that happens is the connection gets closed. There probably is an error message sent from the server, but the connection is already closed so it never bubbles up. Not sure if this is great.

src/Tests/IntegrationTests/TestJetStream.cs
```diff
@@ -937,6 +937,70 @@ public void TestMoreCreateSubscriptionErrors()
...
+            string streamName = "stream-max-payload-test";
+            string subject = "mptest";
+            
+            Options opts = ConnectionFactory.GetDefaultOptions();
+            opts.AllowReconnect = false;
+            NATSServer.QuietOptionsModifier.Invoke(opts);
+
+            using (NATSServer.CreateJetStreamFast())
+            {
+                Assert.Throws<NATSJetStreamException>(() =>
```
scottf, line 953: Synchronous publishes get an exception.

src/Tests/IntegrationTests/TestJetStream.cs
```diff
@@ -937,6 +937,70 @@ public void TestMoreCreateSubscriptionErrors()
...
+                    long size = 0;
+                    for (int x = -1; x < 10; x++)
+                    {
+                        size = 1000 + x;
+                        js.PublishAsync("mptest", new byte[size]);
+                        Thread.Sleep(100);
+                    }
+                }
+                Assert.True(receivedDisconnect);
+            }
```
scottf, line 1003: Asynchronous publishes do not an exception, but the connection is closed.

src/Tests/IntegrationTests/TestBasic.cs
ColinSullivan1: A different issue, but I'm wondering there is a way to pull the remaining data from the 1/2 closed TCP socket.  Doubtful, but it'd be interesting to look into for error conditions like this.

ColinSullivan1 (approved):
LGTM.

PR #795, opened 2023-07-11
[BUG] Missed inheriting code caused failure

scottf:
When refactoring pull sync subscription class to have shared base abstract code in common with the push subscription, the push sync subscription class was copied to the abstract class, but was not changed to inherit from it. This caused calls to `sub.GetConsumerInfo` to fail, since it didn't properly have the consumer name.
This affects all Push subscriptions if the dev calls `sub.GetConsumerInfo`, thankfully that's fairly uncommon. But...KeyValue watches use push under the covers and immediately call `sub.GetConsumerInfo`, so immediately fail.
Sigh. Sorry. It's fixed now. I'll release 1.0.7 as soon as possible.


mtmk (approved):
LGTM

PR #802, opened 2023-07-26
Simplified Ordered Consumer

scottf:


src/NATS.Client/JetStream/IConsumerContext.cs
```diff
@@ -42,68 +42,48 @@ public interface IConsumerContext
...
         /// </summary>
-        /// <param name="maxWaitMillis">the max wait value in milliseconds</param>
+        /// <param name="maxWaitMillis">optional max wait value in milliseconds. Defaults to {@value BaseConsumeOptions#DEFAULT_EXPIRES_IN_MS}</param>
         /// <returns>the next message or null if the max wait expires</returns>
-        Msg Next(int maxWaitMillis);
+        Msg Next(int maxWaitMillis = BaseConsumeOptions.DefaultExpiresInMillis);
 
         /// <summary>
-        /// Create a one use Fetch Consumer using all defaults other than the number of messages. See {@link IFetchConsumer}
+        /// Create a one use Fetch Consumer using all defaults other than the number of messages. See IFetchConsumer
```
mtmk:
```suggestion
        /// Create a one use Fetch Consumer using all defaults other than the number of messages. See <see cref="IFetchConsumer"/>
```

src/NATS.Client/JetStream/IStreamContext.cs
```diff
@@ -143,5 +144,23 @@ public interface IStreamContext
...
         /// <returns>true if the delete succeeded</returns>
         bool DeleteMessage(ulong seq, bool erase);
+        
+        /// <summary>
+        /// Create a long-running Ordered IIterableConsumer with custom ConsumeOptions. See IIterableConsumer and ConsumeOptions
+        /// IIterableConsumer requires the developer call nextMessage.  
+        /// </summary>
+        /// <param name="config">the configuration for the ordered consumer</param>
+        /// <param name="consumeOptions">optional custom consume options</param>
+        /// <returns></returns>
```
mtmk:
```suggestion
        /// <returns><see cref="IIterableConsumer"/></returns>
```

src/NATS.Client/JetStream/IStreamContext.cs
```diff
@@ -143,5 +144,23 @@ public interface IStreamContext
...
+        IIterableConsumer OrderedConsume(OrderedConsumerConfiguration config, ConsumeOptions consumeOptions = null);
+
+        /// <summary>
+        /// Create a long-running Ordered IMessageConsumer with custom ConsumeOptions. See IMessageConsumer and ConsumeOptions
+        /// </summary>
+        /// <param name="config">the configuration for the ordered consumer</param>
+        /// <param name="handler">the EventHandler used for receiving messages</param>
+        /// <param name="consumeOptions">optional custom consume options</param>
+        /// <returns></returns>
```
mtmk:
```suggestion
        /// <returns><see cref="IMessageConsumer"/></returns>
```

src/NATS.Client/JetStream/MessageConsumerBase.cs
```diff
@@ -26,15 +26,21 @@ internal class MessageConsumerBase : IMessageConsumer
         protected PullMessageManager pmm;
         protected JetStreamPullApiImpl pullImpl;
         protected bool stopped;
+        protected ConsumerInfo lastConsumerInfo;
```
mtmk:
```suggestion
        protected volatile ConsumerInfo lastConsumerInfo;
```
Seen it in `lock`. This would make it thread safe for reads.

src/NATS.Client/JetStream/MessageConsumerBase.cs
```diff
@@ -26,15 +26,21 @@ internal class MessageConsumerBase : IMessageConsumer
...
             subLock = new object();
+            this.lastConsumerInfo = lastConsumerInfo;
         }
 
         protected void InitSub(IJetStreamSubscription inSub)
         {
             sub = inSub;
+            if (lastConsumerInfo == null)
+            {
+                lastConsumerInfo = sub.GetConsumerInformation();
```
mtmk:
```suggestion
                lock (subLock)
                    if (lastConsumerInfo == null)
                        lastConsumerInfo = sub.GetConsumerInformation();
```
This is the double-check lock. It is a bit confusing. If you prefer just lock both read and write access to `lastConsumerInfo`.

src/NATS.Client/JetStream/StreamContext.cs
```diff
@@ -116,5 +115,45 @@ public bool DeleteMessage(ulong seq, bool erase)
...
+        public IIterableConsumer OrderedConsume(OrderedConsumerConfiguration config, ConsumeOptions consumeOptions = null)
+        {
+            Validator.Required(config, "Ordered Consumer Config");
+            ConsumerConfiguration cc = GetBackingConsumerConfiguration(config);
+            PullSubscribeOptions pso = new OrderedPullSubscribeOptionsBuilder(StreamName, cc).Build();
+            return new IterableConsumer(new SubscriptionMaker(js, pso, cc.FilterSubject), 
+                consumeOptions == null ? ConsumeOptions.DefaultConsumeOptions : consumeOptions, null);
+        }
+
+        public IMessageConsumer OrderedConsume(OrderedConsumerConfiguration config, EventHandler<MsgHandlerEventArgs> handler,
```
mtmk: `OrderedConsume` overloads seems to be returning different things. Would it be confusing for the dev? Feel free to ignore this comment. It might be me.

mtmk (changes requested):

src/NATS.Client/JetStream/StreamContext.cs
scottf: Not sure what to do here. I'll ask the team

src/NATS.Client/JetStream/IConsumerContext.cs
scottf: fixed

src/NATS.Client/JetStream/IStreamContext.cs
scottf: fixed

src/NATS.Client/JetStream/IStreamContext.cs
scottf: fixed

src/NATS.Client/JetStream/MessageConsumerBase.cs
scottf: discussed

mtmk (approved):
LGTM

mtmk (approved):

mtmk (approved):

mtmk (approved):

mtmk (approved):

src/NATS.Client/JetStream/ConsumerContext.cs
```diff
@@ -51,13 +128,40 @@ public ConsumerInfo GetCachedConsumerInfo()
...
+        public Msg Next(int maxWaitMillis = DefaultExpiresInMillis)
         {
-            if (maxWaitMillis < MinExpiresMills) 
+            lock (stateLock)
             {
-                throw new ArgumentException($"Max wait must be at least {MinExpiresMills} milliseconds.");
+                CheckState();
+                if (maxWaitMillis < MinExpiresMills) 
+                {
+                    throw new ArgumentException($"Max wait must be at least {MinExpiresMills} milliseconds.");
```
mtmk, line 138: Should we fix it instead of blowing up? i.e. set maxWait to MinExpires quietly, maybe with a warning log

src/NATS.Client/JetStream/FetchConsumer.cs
```diff
@@ -21,26 +21,53 @@ namespace NATS.Client.JetStream
...
+            maxWaitMillis = fetchConsumeOptions.ExpiresInMillis;
             PullRequestOptions pro = PullRequestOptions.Builder(fetchConsumeOptions.MaxMessages)
                 .WithMaxBytes(fetchConsumeOptions.MaxBytes)
-                .WithExpiresIn(fetchConsumeOptions.ExpiresIn)
+                .WithExpiresIn(fetchConsumeOptions.ExpiresInMillis)
                 .WithIdleHeartbeat(fetchConsumeOptions.IdleHeartbeat)
                 .Build();
-            ((JetStreamPullSubscription)sub).pullImpl.Pull(pro, false, null);
+            InitSub(subscriptionMaker.Subscribe());
+            pullSubject = ((JetStreamPullSubscription)sub).pullImpl.Pull(pro, false, null);
```
mtmk, line 39: Is there a chance of invalid cast here? Is there a way to change the type hierarchy to avoid casts?
It's an internal interface so no need to be defensive (i.e. checking types or using `as` or something) so feel free to ignore this comment as long as we're certain that cast will succeed.

src/NATS.Client/JetStream/FetchConsumer.cs
```diff
@@ -21,26 +21,53 @@ namespace NATS.Client.JetStream
...
         }
 
         public Msg NextMessage()
```
mtmk, line 42: Can we make return type nullable and possible document that it may return `null`?

src/NATS.Client/JetStream/FetchConsumer.cs
```diff
@@ -52,18 +79,21 @@ public Msg NextMessage()
...
-                // if the manager thinks it has received everything in the pull, it means
-                // that all the messages are already in the internal queue and there is
-                // no waiting necessary
-                if (timeLeftMillis < 1 || pmm.pendingMessages < 1 || (pmm.trackingBytes && pmm.pendingBytes < 1))
+                // if the timer has run out, don't allow waiting
+                // this might happen once, but it should already be noMorePending
+                if (timeLeftMillis < 1)
                 {
-                    return ((JetStreamPullSubscription)sub).NextMessage(1); // 1 is the shortest time I can give
+                    return ((JetStreamPullSubscription)sub)._nextUnmanagedNoWait(pullSubject); // 1 is the shortest time I can give
```
mtmk, line 86: Public members shouldn't start with underscore, unless I'm missing something here.

src/NATS.Client/JetStream/IBaseConsumerContext.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// SIMPLIFICATION IS EXPERIMENTAL AND SUBJECT TO CHANGE
+    /// </summary>
+    public interface IBaseConsumerContext
+    {
+        /// <summary>
+        /// Read the next message with optional provided max wait
+        /// </summary>
+        /// <param name="maxWaitMillis">optional max wait value in milliseconds. Defaults to {@value BaseConsumeOptions#DEFAULT_EXPIRES_IN_MS}</param>
+        /// <returns>the next message or null if the max wait expires</returns>
+        Msg Next(int maxWaitMillis = BaseConsumeOptions.DefaultExpiresInMillis);
```
mtmk, line 16: Can we return nullable?

src/Samples/SimplificationIterableConsumer/IterableConsumerExample.cs
```diff
@@ -79,23 +79,26 @@ public static void Main(String[] args)
...
 
                             Console.WriteLine("Starting post-stop loop.");
                             msg = consumer.NextMessage(1000);
                             while (msg != null)
                             {
                                 msg.Ack();
-                                report("Post-stop loop running", sw.ElapsedMilliseconds, ++count);
+                                if (++count % REPORT_EVERY == 0)
+                                {
+                                    report("Post-stop loop running", sw.ElapsedMilliseconds, ++count);
```
mtmk, line 96: Incrementing count twice. Is that intentional?

src/Samples/SimplificationMessageConsumer/MessageConsumerExample.cs
```diff
@@ -72,13 +72,18 @@ public static void Main(String[] args)
...
                     latch.Wait();
+                    
                     // once the consumer is stopped, the client will drain messages
                     Console.WriteLine("Stop the consumer...");
-                    consumer.Stop(1000);
-                    Thread.Sleep(1000); // enough for messages to drain after stop
+                    consumer.Stop();
+                    
+                    // wait until the consumer is finished
+                    while (!consumer.Finished) {
```
mtmk, line 84: I know it's just an example but if this is an expected usage pattern `consumer.Finished` must be thread safe.

mtmk (commented):
I think there are a couple of comments might require change. It's mostly suggestions.

src/NATS.Client/JetStream/ConsumerContext.cs
scottf: No, this is a developer coding error. We don't want the user to get behavior that they did not expect. This is a pretty important theme in general in the clients. Let them fail. I will make sure the doc is there.

src/NATS.Client/JetStream/FetchConsumer.cs
scottf: Unfortunately the hierarchy predates me and was designed this way, I'm working around things as it is. IJetStreamSubscription is the best common base. I agree it's not perfect. I will will make a review pass and double check if there is something better higher I can make "sub"

src/NATS.Client/JetStream/FetchConsumer.cs
scottf: Not sure what variable you are talking about or what return.

scottf: Yes, I will go through and update in a new PR.

src/NATS.Client/JetStream/FetchConsumer.cs
scottf: That's not a public member, its internal. I could change it to NextUnmanageNoWaitInternal or something. Suggestions?

src/NATS.Client/JetStream/IBaseConsumerContext.cs
scottf: yes will change in another PR

src/Samples/SimplificationIterableConsumer/IterableConsumerExample.cs
scottf: That's a bug, will fix.

src/Samples/SimplificationMessageConsumer/MessageConsumerExample.cs
scottf:
```
public bool Finished { get; internal set; }
```
Not sure what the problem is to just read this variable. It's should not be a problem if it gets changed from false to true in the middle of a read and the user loops one more time.

I could use Interlocked and a backing variable if you think it's critical.

src/NATS.Client/JetStream/IBaseConsumerContext.cs
scottf: Cannot change. Nullable reference types are not available.
image

src/NATS.Client/JetStream/FetchConsumer.cs
scottf: Since it's internal I can change it at any time

src/NATS.Client/JetStream/FetchConsumer.cs
scottf: See comment below

src/NATS.Client/JetStream/IBaseConsumerContext.cs
mtmk: Sorry, didn't realise the project settings i.e. c# 8.0 support. It's a nice to have. We might be able to introduce it later maybe.

src/Samples/SimplificationMessageConsumer/MessageConsumerExample.cs
mtmk: A volatile backing field should do it. Assuming setting and getting/checking are done from different threads, the problem is the checking thread might not see the update at all and stay in while loop forever even though the thing is finished.

PR #808, opened 2023-08-15
Fix use of JsOptions timeout

scottf:
When making requests from JetStream classes, the JetStreamOptions. There was use of the Connection options timeout instead in one place and this was fixed. Also, the default JetStreamOption timeout was resolved to the connection timeout if the JetStreamOption timeout was not explicitly set.


mtmk (approved):
LGTM

PR #809, opened 2023-08-17
Service / Simplification Documentation and Release Prep

scottf:
* Documented Service Framework
* Removed experimental tags from Service and Simplification
* Fixed `Dictionary` to `IDictionary` return values
* Fixed `String` to `string`


arondi (approved):
LGTM

PR #813, opened 2023-08-29
stream/consumer info timestamps, stream config first_seq

scottf:
Server version 2.10 provides 
* a timestamp for when the stream or consumer info was gathered
* the ability to set the first sequence number for a stream


src/Tests/UnitTests/TestBase.cs
```diff
@@ -120,87 +121,91 @@ public string[] SplitLines(string text)
         public const string KEY = "key";
         public const string DATA = "data";
         
-        public static string Stream(object seq)
```
scottf, line 123: THIS WHOLE CLASS IS JUST TEST SCAFOLDING. Nothing to see here!

src/NATS.Client/JetStream/StreamConfiguration.cs
```diff
@@ -193,42 +197,46 @@ public sealed class StreamConfigurationBuilder
             internal bool _denyPurge;
             internal bool _discardNewPerSubject;
             internal Dictionary<string, string> _metadata = new Dictionary<string, string>();
+            internal ulong _firstSequence;
 
             public StreamConfigurationBuilder() {}
             
             public StreamConfigurationBuilder(StreamConfiguration sc)
             {
-                if (sc == null) return;
```
scottf, line 201: Just some cleanup.

piotrpio (approved):
LGTM

PR #821, opened 2023-09-24
Multiple Filter Subjects and Subject Validation

scottf:


src/NATS.Client/Internals/Validator.cs
```diff
@@ -45,30 +45,79 @@ internal static class Validator
...
+            if (subject.EndsWith(".")) {
+                return new Tuple<bool, string>(false, $"{label} cannot end with '.'");
             }
+
             string[] segments = subject.Split('.');
-            for (int x = 0; x < segments.Length; x++) {
-                string segment = segments[x];
-                if (segment.Equals(">")) {
-                    if (cantEndWithGt || x != segments.Length - 1) { // if it can end with gt, gt must be last segment
-                        throw new ArgumentException(label + " cannot contain '>'");
```
mtmk, line 65: I might be missing something but check for '>' being in the middle seems like not being done below.

src/NATS.Client/Internals/Validator.cs
```diff
@@ -678,6 +742,26 @@ public static bool SequenceEqual<TSource>(IList<TSource> l1, IList<TSource> l2,
...
+                return l2 == null || l2.Count == 0;
+            }
+
+            if (l2 == null || l1.Count != l2.Count)
+            {
+                return false;
+            }
+
+            foreach (T t in l1) {
+                if (!l2.Contains(t)) {
```
mtmk, line 765: I don't think this does what it says in the method name [1,2,2] != [1,1,2]

src/NATS.Client/JetStream/ConsumerConfiguration.cs
```diff
@@ -526,7 +557,51 @@ public ConsumerConfigurationBuilder WithMaxDeliver(long? maxDeliver)
             /// <returns>The ConsumerConfigurationBuilder</returns>
             public ConsumerConfigurationBuilder WithFilterSubject(string filterSubject)
             {
-                _filterSubject = EmptyAsNull(filterSubject);
+                if (string.IsNullOrWhiteSpace(filterSubject)) 
+                {
+                    this._filterSubjects = null;
+                }
+                else {
+                    this._filterSubjects = new List<string>();
```
mtmk, line 566: Can this be called more than once?

mtmk (changes requested):
looks good in general only one or two minor points need clarification.

src/NATS.Client/Internals/Validator.cs
scottf: @mtmk see Line 99

src/NATS.Client/Internals/Validator.cs
scottf: I probably need a better name for this and a note. Since it is comparing lists of subject filters, there will never be duplicates

src/NATS.Client/JetStream/ConsumerConfiguration.cs
scottf: Yes. It's in a builder. This is akin to set/put versus add. I can't change the api name though, I'll just doc it better.

src/NATS.Client/Internals/Validator.cs
scottf:
```
// This function tests filter subject equivalency
// It does not care what order and also assumes that there are no duplicates.
// From the server: consumer subject filters cannot overlap [10138]
public static <T> boolean consumerFilterSubjectsAreEquivalent(List<T> l1, List<T> l2)
```

src/NATS.Client/JetStream/ConsumerConfiguration.cs
scottf: Added api doc comment
> Replaces any other filter subjects set in the builder

src/NATS.Client/Internals/Validator.cs
mtmk: @scottf ah actually I saw that one. Isn't that checking whether `>` or `*` appearing with in a single token/segment? e.g. `foo.b>ar` I was wondering if you'd need to flag `foo.>.bar` as error as well. (Unless I'm running the function wrongly in my head, in that case please ignore)

scottf: Got it `foo.*.bar` is okay but not `foo.>.bar`

scottf: fixed

mtmk (approved):
LGTM

PR #824, opened 2023-10-11
[Service] Custom endpoint queue support and fixed InfoResponse to show proper subject.

scottf:


src/NATS.Client/Service/Endpoint.cs
```diff
@@ -31,39 +36,54 @@ public class Endpoint : JsonSerializable
...
 
         /// <summary>
         /// Directly construct an Endpoint with a name, the subject, and metadata
         /// </summary>
         /// <param name="name">the name</param>
         /// <param name="subject">the subject</param>
+        /// <param name="queueGroup">the queue group</param>
         /// <param name="metadata">the metadata</param>
-        public Endpoint(string name, string subject, IDictionary<string, string> metadata) : this(name, subject, metadata, true) {}
+        public Endpoint(string name, string subject, string queueGroup, IDictionary<string, string> metadata) : this(name, subject, null, metadata, true) {}
```
mtmk: Need to pass queueGroup here, not null?

scottf: Yes. Fixed.

scottf: I need more tests!

mtmk (approved):
LGTM

src/NATS.Client/Service/Endpoint.cs
mtmk: Just a thought. You can't use the default param value syntax here because the target lang is old, can you? e.g. `..., string queueGroup = null,...`

scottf: I might be able to. I've only done it for all the last parameters but that may just be my lack of knowledge. It's not really that important, there is a builder. I had made the constructors when there were just a few parameters and then it got out of hand with changes and I can't pull them now

PR #830, opened 2023-10-18
Direct Message Subject Header May Contain Multiple Subjects

scottf:


src/NATS.Client/MsgHeader.cs
```diff
@@ -249,6 +249,30 @@ public string[] GetValues(string name)
...
+        /// <summary>
+        /// Gets the last value for the specific key.
+        /// Will be null if the key is not found
+        /// </summary>
+        /// <param name="name"></param>
+        /// <returns></returns>
+        public string GetLast(string name)
+        {
+            string[] all = nvc.GetValues(name);
+            return all == null ? null: all[all.Length-1];
```
mtmk, line 273: Length > 0 would be nice

mtmk: I guess getvalues might return null in that case, so maybe not needed.

mtmk (changes requested):
Looks good. Just wanted to check if a bounds check is necessary.

src/NATS.Client/MsgHeader.cs
scottf: I just did a test to verify. The backing NameValueCollection returns null if the key is not present so GetValues returns null as well.

mtmk (approved):
LGTM

PR #844, opened 2023-11-23
Expose pending limits in the Options

LaurensVergote:
Had a use case where I wanted to configure these on a connection-based level and noticed these options were not exposed yet.
In theory, Options.SubChannelLength can be obsoleted now, but might want to keep it for backwards compatibility


sixlettervariables (approved):
Consider an obsolete attribute on the old property if it now unused. I think the move from int32 to int64 is fine, but would be helped by that note.

LaurensVergote:
> Consider an obsolete attribute on the old property if it now unused. I think the move from int32 to int64 is fine, but would be helped by that note.

Marked as obsolete

PR #850, opened 2023-12-18
Avoid Port Lock when reconnecting with a localhost nats-server

### LINKED ISSUE #847: Possible port lock situation when trying to connect to a local, not running, NATS server
### Observed behavior

When creating a connection to a [local hostname](https://learn.microsoft.com/en-us/dotnet/api/system.net.dns.gethostname?view=net-8.0), it's possible that the [TcpClient created](https://github.com/nats-io/nats.net/blob/main/src/NATS.Client/Connection.cs#L386) will grab the target port as its local port. This seems to result in a connection successfully being established and the client application grabbing the port, making it impossible for the nats server to ever start up later on using that port.

output of netstat (table format for readability):
>netstat -abon | findstr 4222

| Proto | Local Address | Foreign Address | State | PID |
| -- | -- | -- | -- | -- |
| TCP | 10.3.1.54:4222 | 10.3.1.54:4222 | ESTABLISHED | 4708 |

Possible solutions:
- Use the TcpClient constructor with IPEndpoint input ([TcpClient](https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.tcpclient.-ctor?view=net-8.0#system-net-sockets-tcpclient-ctor(system-net-ipendpoint)))
- After constructing the TcpClient, modify the internal socket ([TcpClient.Client](https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.tcpclient.client?view=net-8.0)) to exclude the target port from being used as the local endpoint's port

Comments:
While I'm not 100% certain on the root cause of this, the above description is my best educated guess on why this would happen.
I've only noticed it on dev machines that run the host process, which tries to connect every 1 second to NATS, and while the target local nats-server, is not running.

This results in not being able to start the local nats-server anymore.

I know it's very niche but it's better if its noted down at least, even if there's no intent to fix it or spend time on it.

### Expected behavior

The NATS client does not use the target endpoint's port as the local endpoint's when connecting with a local endpoint

### Server and client version

Server version not relevant.
client version: 1.1.1

### Host environment

Running on Windows 10 Enterprise, loading the NATS client library as .netstandard 2.0 from a .NET 6.0 host process.

### Steps to reproduce

Continually call Connect until you randomly get the target's endpoint port as your local endpoint's port.

LaurensVergote:
Fixes #847

### DIFF
--- a/src/NATS.Client/Connection.cs
+++ b/src/NATS.Client/Connection.cs
@@ -400,6 +400,12 @@ namespace NATS.Client
                         client = null;
                         throw new NATSConnectionException("timeout");
                     }
+                    if (HasBoundTargetPort(client, s))
+                    {
+                        close(client);
+                        client = null;
+                        throw new NATSConnectionException("wrong port bound");
+                    }
 
                     client.NoDelay = false;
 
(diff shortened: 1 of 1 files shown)


LaurensVergote:
Reproduction using just the nats.client library: https://gist.github.com/LaurensVergote/ba4b4ea1825f2ff5fed3d1aee9214dd8
Just make sure a nats-server.exe is available next to the compiled binary

PR #853, opened 2024-01-15
Auth Expiration Resubscribe

scottf:
When an auth callout jwt causes an expiration, don't close subscriptions so they can be re-subscribed on reconnect.


mtmk (approved):
LGTM

PR #854, opened 2024-01-16
Reconnect on connect, set state

scottf:
The reconnect loop exits early when status is closed. The first connect sets status to close when it can't connect, so reconnect exited early.
The fix is that before going to reconnect, change the status to disconnected, like it would be if it had been connected then wasn't.


aricart (approved):
LGTM

PR #859, opened 2024-01-31
Improve parseError message to include position and UTF8 buffer string

Madgvox:
`byte[]`'s `ToString` prints `System.Byte[]`. This is not helpful when diagnosing the state of the parser when an error occurs, so I instead print the UTF8 encoded string version of the buffer. I also print the position for good measure.

### DIFF
--- a/src/NATS.Client/Parser.cs
+++ b/src/NATS.Client/Parser.cs
@@ -13,6 +13,7 @@
 
 using System;
 using System.IO;
+using System.Text;
      
 namespace NATS.Client
 {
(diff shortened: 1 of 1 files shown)


scottf (changes requested):
This is a good idea, but we cannot really print the buffer, it might not even be a string and it could be long.
Maybe, printing the position and length?
Also, since it's private, we can change it's signature and give more information about where it's called from, like which operation it was parsing, or even some custom information from each place it's called.

scottf:
@Madgvox If this isn't going to move forward I'm going to close it without merging.

PR #862, opened 2024-02-07
Updated MinExpiresMills to 100

tbazen:
**MinExpiresMills** is the least possible time to wait value for fetch and consume, and currently it is set to 1 second. It would be great if either the allowed value is lower than the current value or if positive integer is acceptable.


scottf:
@bazen-teklehaymanot This value was discussed by the client development team and documented here: https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-37.md#defaults-and-constraints
I cannot change this value without agreement from the team. My suggestion is to file an issue in that repo and reference the part in ADR-37. I would include some reason and an example use case to help defend why you want to make this change.

PR #872, opened 2024-03-05
KV Atomic Delete and Purge

scottf:
https://github.com/nats-io/nats-architecture-and-design/issues/271
https://github.com/nats-io/nats.java/issues/1091

#### What's in this patch?

1. New public methods in IKeyValue for delete and purge with optimistic concurrency support:
```
void Delete(String key, ulong expectedRevision);
void Purge(String key, ulong expectedRevision);
```

2. New unit test `TestAtomicDeletePurge` with positive and negative cases.


mtmk (approved):
LGTM

Issue #880, opened 2024-03-21
Heartbeat messages cause slow consumer


ideahitme-github:
### Observed behavior

When a subscription with heartbeat enabled has been running for a while ( few days) it is eventually guaranteed to lead to a slow consumer exception. I believe the reason for this is that every heartbeat received increases the pending message count and that never gets decremented. Eventually it will exceed the pending message limit and when a new (non heartbeat) message is received it will throw slow consumer exception

### Expected behavior

Subscription can stay alive as long as it wants without leading to a slow consumer exception

### Server and client version

Client v.1.0.4 (but I believe the problem exists in current version as well)
Server v.2.9.16

### Steps to reproduce

Create a subscription e.g. 

```
        var options = PushSubscribeOptions.Builder()
            .WithStream(_clientContext.PersistentStreamName)
            .WithPendingMessageLimit(10)
            .WithConfiguration(
                ConsumerConfiguration.Builder()
                    .WithIdleHeartbeat(6000)
                    .Build())
            .Build(); 
            ...
        js.PushSubscribeAsync(subject: SubjectName, handler: Handler, autoAck: true, options);
```

Wait couple of minutes and then push a message into the jetstream and see `AsyncErrorEventHandler` being triggered with `Slow Consumer` exception

ideahitme-github:
Maybe something like this would be a good fix: 

https://github.com/nats-io/nats.net/compare/main...ideahitme-github:nats.net:patch-1

scottf:
@ideahitme-github I think I should not count heartbeats at all. Take a look at this. https://github.com/nats-io/nats.net/pull/881/files
Heartbeat messages get a false from this
```
BeforeChannelAddCheck.Invoke(msg)
```
so I'm not going to even count them.

ideahitme-github:
that works too. Thanks @scottf

PR #881, opened 2024-03-22
Messages not added to the channel don't count as pending for client side slow consumer purposes

scottf:
In response to https://github.com/nats-io/nats.net/issues/880


mtmk (approved):
LGTM

PR #885, opened 2024-04-08
Ability to set user credentials from strings(s)

scottf:


src/NATS.Client/BaseUserJWTHandler.cs
```diff
@@ -0,0 +1,81 @@
...
+    /// <see cref="Options.UserSignatureEventHandler"/>.  This class is
+    /// not normally used directly, but is provided to extend or use for
+    /// utility methods to read a private seed or user JWT.
+    /// </summary>
+    public class BaseUserJWTHandler
+    {
+        protected static string _loadUser(string text)
+        {
+            StringReader reader = null;
+            try
```
mtmk: we could use `using` here but I see it's lifted from the original file, so better to leave as is.

mtmk (approved):
LGTM

src/NATS.Client/BaseUserJWTHandler.cs
```diff
@@ -0,0 +1,81 @@
...
+                string text = secure.ToString();
+                
+                // if it's a nk file, it only has the nkey
+                if (text.StartsWith("SU"))
+                {
+                    return Nkeys.FromSeed(text);
+                }
+
+                // otherwise assume it's a creds file.
+                reader = new StringReader(text);
```
sixlettervariables: Not sure why the original code had the `ToString`
```suggestion
                // if it's a nk file, it only has the nkey
                if (secure.StartsWith("SU"))
                {
                    return Nkeys.FromSeed(secure);
                }

                // otherwise assume it's a creds file.
                reader = new StringReader(secure);
```

src/NATS.Client/DefaultUserJWTHandler.cs
```diff
@@ -22,7 +23,7 @@ namespace NATS.Client
     /// not normally used directly, but is provided to extend or use for
     /// utility methods to read a private seed or user JWT.
     /// </summary>
-    public class DefaultUserJWTHandler
+    public class DefaultUserJWTHandler : BaseUserJWTHandler
```
sixlettervariables: Since we had a breaking change earlier I went and double checked this would not introduce any issues, and I don't believe it will:
> ❓ REQUIRES JUDGMENT: Introducing a new base class

> A type can be introduced into a hierarchy between two existing types if it doesn't introduce any new [abstract](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/abstract) members or change the semantics or behavior of existing types. For example, in .NET Framework 2.0, the [DbConnection](https://learn.microsoft.com/en-us/dotnet/api/system.data.common.dbconnection) class became a new base class for [SqlConnection](https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlconnection), which had previously derived directly from [Component](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.component).

https://learn.microsoft.com/en-us/dotnet/core/compatibility/library-change-rules#modifications-to-the-public-contract

src/NATS.Client/DefaultUserJWTHandler.cs
```diff
@@ -56,33 +57,19 @@ public DefaultUserJWTHandler(string jwtFilePath, string credsFilePath)
...
-            finally
+            string user = _loadUser(text);
+            if (user == null)
             {
-                Nkeys.Wipe(text);
-                Nkeys.Wipe(line);
-                reader?.Dispose();
+                throw new NATSException("Credentials file does not contain a JWT");
             }
+            return user.ToString();
```
sixlettervariables:
```suggestion
            return user;
```

sixlettervariables (approved):
Minor nits; also double checked we wouldn't introduce any issues with the new base class.

src/NATS.Client/DefaultUserJWTHandler.cs
scottf: I can change from base to a delegate

src/NATS.Client/DefaultUserJWTHandler.cs
scottf: yeah, that was left over from trying secure.

src/NATS.Client/BaseUserJWTHandler.cs
scottf: leftover from when I was trying secure. Will fix.

src/NATS.Client/DefaultUserJWTHandler.cs
```diff
@@ -91,47 +78,23 @@ public static string LoadUserFromFile(string path)
...
-                else
-                    return kp;
+                }
+                return kp;
             }
             finally
             {
-                Nkeys.Wipe(line);
-                Nkeys.Wipe(text);
-                Nkeys.Wipe(seed);
```
scottf, line 134: FYI, these were removed because the wipe tries to wipe a string which is not possible. The wipe method was recently changed to a no-op.

mtmk (approved):

sixlettervariables (approved):

PR #888, opened 2024-04-29
Fix line reader to resolve buffer issues on TLS connections

### LINKED ISSUE #586: NotSupportedException in Connection.publish getting BufferedStream.Position
We are seeing a `System.NotSupportedException` with the following stack trace:
```
   at System.IO.__Error.SeekNotSupported()
   at System.IO.BufferedStream.get_Position()
   at NATS.Client.Connection.publish(String subject, String reply, Byte[] headers, Byte[] data, Int32 offset, Int32 count, Boolean flushBuffer)
   at NATS.Client.Connection.<requestAsync>d__141.MoveNext()
```

It happens intermittently and isn't 100% reproducible, but here is my best guess as to what is happening:

In the C# class `Connection` there is a field `private Stream bw` which is a `BufferedStream` over another underlying `Stream`.

In the method `publish` if the `status` is `ConnState.RECONNECTING` then the getter for the property `bw.Position` is called.  This getter is only supported if the underlying `Stream` supports seeking.  As we can see from the stack trace above, in that case, the underlying stream does not support seeking, and this causes a `NotSupportedException` to be thrown.

Here is my best guess as to what happened:

When the connection first enters `ConnState.RECONNECTING`, `bw` is created over a `MemoryStream` called `pending` which does support seeking and so is not the cause.

However, later on, the method `doReconnect` calls the method `createConn` which changes the field `bw` to now be created over a stream associated with the TCP socket (which does not support seeking).  After `createConn` has returned successfully, `doReconnect` then calls `processConnectInit`.  As a thought experiment, consider what happens if an exception is thrown by `processConnectInit`: it will be caught in the `catch` block in `doReconnect`, and the `status` will be set to `ConnState.RECONNECTING` but it will not change `bw` to be over a `MemoryStream` and so `bw` will continue to have an underlying stream associated with the TCP socket (which does not support seeking).  The `catch` block will then `continue` and it will then select the next server and release the `mu` lock before it calls the `ReconnectDelayHandler`.  At this point imagine that another thread calls `publish`.  `publish` will throw the stack trace above, because `mu` is not locked, and the `status` is `ConnState.RECONNECTING`, and the underlying stream of `bw` does not support seeking.

I think the fix is that in every place where the `status` is set to be `ConnState.RECONNECTING`, `bw` should be reset to be created over the `MemoryStream` called `pending`

mtmk:
The StreamReader usage when reading the server INFO causes buffer issues on the stream, as it is also read in the reader loop. The StreamReader might consume or return the read bytes back, leading to problems when used on a TLS connection with .NET versions lower than 8. This fix removes the StreamReader usage to address these issues.

fixes https://github.com/nats-io/nats.net/issues/586#issuecomment-2076701242

### DIFF
--- a/src/NATS.Client/Connection.cs
+++ b/src/NATS.Client/Connection.cs
@@ -1480,12 +1480,12 @@ namespace NATS.Client
             string result = null;
             try
             {
-                result = br.ReadUntilNewline();
+                result = br.ReadUntilCrlf();
 
                 // If opts.verbose is set, handle +OK.
                 if (opts.Verbose && IC.okProtoNoCRLF.Equals(result))
                 {
-                    result = br.ReadUntilNewline();
+                    result = br.ReadUntilCrlf();
                 }
             }
             catch (Exception ex)
@@ -1518,7 +1518,7 @@ namespace NATS.Client
 
         private Control readOp()
         {
-            return new Control(br.ReadUntilNewline());
+            return new Control(br.ReadUntilCrlf());
         }
 
         private void processDisconnect()
@@ -5322,7 +5322,7 @@ namespace NATS.Client
         /// The performance of this reading approach is relatively low, so it should only
         /// be used in scenarios with low traffic, such as reading the server INFO.
         /// </summary>
-        internal static string ReadUntilNewline(this Stream stream)
+        internal static string ReadUntilCrlf(this Stream stream)
         {
             const int maxAllowedSize = MaxControlLineSize * 16;
             var buffer = new byte[MaxControlLineSize];
@@ -5362,7 +5362,12 @@ namespace NATS.Client
                 }
             }
 
-            return Encoding.UTF8.GetString(buffer, 0, read - (foundCR ? 0 : 1));
+            if (!foundCR)
+            {
+                throw new NATSProtocolException("Control line does not end with CRLF.");
+            }
+            
+            return Encoding.UTF8.GetString(buffer, 0, read - 1);
         }
     }
 }
--- a/src/NATS.Client/Connection.cs
+++ b/src/NATS.Client/Connection.cs
@@ -5324,7 +5324,7 @@ namespace NATS.Client
         /// </summary>
         internal static string ReadUntilNewline(this Stream stream)
         {
-            const int maxAllowedSize = MaxControlLineSize * 1024;
+            const int maxAllowedSize = MaxControlLineSize * 16;
             var buffer = new byte[MaxControlLineSize];
             int byteValue;
             bool foundCR = false;
--- a/src/NATS.Client/Connection.cs
+++ b/src/NATS.Client/Connection.cs
@@ -1478,28 +1478,19 @@ namespace NATS.Client
             }
 
             string result = null;
-            StreamReader sr = null;
             try
             {
-                // TODO:  Make this reader (or future equivalent) unbounded.
-                // we need the underlying stream, so leave it open.
-                sr = new StreamReader(br, Encoding.UTF8, false, MaxControlLineSize, true);
-                result = sr.ReadLine();
+                result = br.ReadUntilNewline();
 
                 // If opts.verbose is set, handle +OK.
                 if (opts.Verbose && IC.okProtoNoCRLF.Equals(result))
                 {
-                    result = sr.ReadLine();
+                    result = br.ReadUntilNewline();
                 }
             }
             catch (Exception ex)
             {
                 throw new NATSConnectionException("Connect read error", ex);
-            }
-            finally
-            {
-                if (sr != null)
-                    sr.Dispose();
             }
 
             if (IC.pongProtoNoCRLF.Equals(result))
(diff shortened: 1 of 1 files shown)


src/NATS.Client/Connection.cs
```diff
@@ -5327,4 +5309,53 @@ public IObjectStoreManagement CreateObjectStoreManagementContext(ObjectStoreOpti
...
+                }
+                else if (byteValue == '\n' && foundCR)
+                {
+                    break;
+                }
+                else
+                {
+                    if (foundCR)
+                    {
+                        buffer[read - 2] = (byte)'\r';
```
sixlettervariables: If the first byte is `\r`, `read` will be 1, and `foundCR` will be `true`. Walking this loop in my head.

mtmk: yea I think it should be -1 good catch 💯

src/NATS.Client/Connection.cs
```diff
@@ -5327,4 +5309,53 @@ public IObjectStoreManagement CreateObjectStoreManagementContext(ObjectStoreOpti
...
+                {
+                    if (foundCR)
+                    {
+                        buffer[read - 1] = (byte)'\r';
+                        foundCR = false;
+                    }
+                    buffer[read] = (byte)byteValue;
+                }
+                
+                if (++read >= MaxControlLineSize)
```
caleblloyd: Previously, `MaxControlLineSize` was being used in [StreamReader(Stream, Encoding, Boolean, Int32, Boolean) ](https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader.-ctor?view=net-8.0#system-io-streamreader-ctor(system-io-stream-system-text-encoding-system-boolean-system-int32-system-boolean)) where it was "The minimum buffer size."

I believe that the Max Control Line Size is technically a Server setting, and that the client is not privy to what that setting is.  If that is the case, we should probably add a buffer here that can grow for the rare case that someone has increased the server's Max Control Line Size

mtmk: good point. made the buffer dynamic.

also kept a size restriction, a bit arbitrary I suppose, wdyt? 

**edit** I don't think we have any limits set on v2 btw

scottf: Max control line notes from the Java api doc...
> Set the maximum length of a control line sent by this connection. This value is also configured in the server but the protocol doesn't currently forward that setting. Configure it here so that the client can ensure that messages are valid before sending to the server.

caleblloyd: 32K or 64K is probably a reasonable size restriction if we want to keep one - that would either be 8x-16x (instead of 1024x) the MaxControlLineSize

But true, I don't think we have a limit on v2 because it would just mean invalid protocol message.  But the size restriction will potentially be helpful by throwing instead of hanging if there is a bunch of gibberish coming down the pipe

mtmk: 👍 set it to 16x (feeling generous)

@scottf does Java client have such a limit when receiving INFO line from the server for instance?

src/NATS.Client/Connection.cs
```diff
@@ -5327,4 +5309,60 @@ public IObjectStoreManagement CreateObjectStoreManagementContext(ObjectStoreOpti
...
+        /// Consumes the stream one byte at a time until a newline is found.
+        /// This approach is used to avoid over-reading or under-reading the stream,
+        /// as it will be accessed in the reader loop as well.
+        /// Using StreamReader instead of this method can cause issues when a line
+        /// is delivered in chunks, especially over global connections, which is more
+        /// likely to occur.
+        /// The performance of this reading approach is relatively low, so it should only
+        /// be used in scenarios with low traffic, such as reading the server INFO.
+        /// </summary>
+        internal static string ReadUntilNewline(this Stream stream)
```
caleblloyd: Maybe call this `ReadUntilCrLf`?  Since "NewLine" can mean many things (cr, lf, or crlf)

src/NATS.Client/Connection.cs
```diff
@@ -5327,4 +5309,60 @@ public IObjectStoreManagement CreateObjectStoreManagementContext(ObjectStoreOpti
...
+                    if (foundCR)
+                    {
+                        buffer[read - 1] = (byte)'\r';
+                        foundCR = false;
+                    }
+                    buffer[read++] = (byte)byteValue;
+                }
+            }
+
+            return Encoding.UTF8.GetString(buffer, 0, read - (foundCR ? 0 : 1));
```
caleblloyd: Another potential edge case - what if the sequence was CR followed by EOF.  `foundCR` was true, but there was no LF.  Should still return the `CR`?

src/NATS.Client/Connection.cs
scottf: Java
```
public static final int DEFAULT_MAX_CONTROL_LINE = 4096;
public static final int DEFAULT_BUFFER_SIZE = 64 * 1024;
```
.NET V1
```
internal const int MaxControlLineSize = 4096;
public const int defaultBufSize = 32768;
```

src/NATS.Client/Connection.cs
caleblloyd: Or should we be throwing if no CRLF was found?  Since the point of the metheod is to detect a CRLF that is required by the protocol?

mtmk: I think EOF (read returning -1) would mean connection is closed or disconnected in which case this would return whatever is read upto that point which probably won't be the complete line or empty string. it should never return the last CR. So maybe an exception is in order.

caleblloyd (approved):
Looks good to me!

Issue #891, opened 2024-05-16
Passing consumer configuration in KV store subscription


MENNAELMESALMY:
### Proposed change

In KV store I want to be able to pass metadata, description, and idleheartbeat time (basically some consumer configuration) to the associated consumer but I can't find a way to pass this config, am I missing something?
if there is currently no way to do it, can we consider adding it?

used version: v1 `1.1.5`

### Use case

Same use cases as having JetStream's consumer metadata, we want to be able to pass some metadata about the current consumers  in the bucket.

### Contribution

I am thinking of having a class similar to `ConsumerConfigurationBuilder` that we can optionally ( with a new constructor) pass to Watch/WatchAll but would like to know if there is a process to follow to have that.

scottf:
I could see having metadata and description because they are purely informational. Things like allowing custom heartbeats would have to be considered by the team and I can run it by them, and if we were going to allow some customization, we really should look at every possible consumer configuration that is available to us considering it's an ordered consumer under the covers.
Alternatively, remember KV is just something on top of JetStream, and all the code is there, so you could make your own watcher implementation, you would have to understand the way subjects are formed.

MENNAELMESALMY:
hi @scottf thanks for the quick reply.
We would like to continue to get any feature that the library provides smoothly so don't want to diverge too much with our own implementation.
If you agree that informational data is harmless (metadata, description) is harmless I could start drafting a PR for that.

For the heartbeat, I understand that the ordered consumer with a push subscription enables control flow ([link](https://github.com/nats-io/nats.net/blob/0fafd5b779fd153f99e4e9c3898f38f03761851d/src/NATS.Client/JetStream/SubscribeOptions.cs#L142-L145)) which enables the heartbeat but since it's not passed I assume that it's always the default I was looking to make it customizable. curious if that can lead to other side effects though, let me know what you and the team think.

PR #894, opened 2024-05-20
pass consumer information in kv store

MENNAELMESALMY:
Implementation for #891


scottf (changes requested):
My initial reaction is that I don't feel good about adding 10 more api. I would be more inclined if you add only 2 more (watch/watch all) with all possible parameters. Also, this has not been approved yet by the client team, I have raised the issue again.

src/NATS.Client/KeyValue/KeyValueConsumerConfiguration.cs
```diff
@@ -0,0 +1,159 @@
...
+
+using System.Collections.Generic;
+using NATS.Client.Internals.SimpleJSON;
+using NATS.Client.JetStream;
+using static NATS.Client.Internals.JsonUtils;
+using static NATS.Client.Internals.Validator;
+
+namespace NATS.Client.KeyValue
+{
+    public sealed class KeyValueConsumerConfiguration : JsonSerializable
```
scottf: Does this have to be JsonSerializable? It's not ever written anywhere, it's just a structure.

src/NATS.Client/KeyValue/KeyValueConsumerConfiguration.cs
```diff
@@ -0,0 +1,159 @@
...
+        {
+            JSONObject o = new JSONObject();
+
+            AddField(o, ApiConstants.Description, Description);
+            AddField(o, ApiConstants.Name, Name);
+            AddField(o, ApiConstants.Metadata, Metadata);
+            return o;
+        }
+
+        internal IList<string> GetChanges(KeyValueConsumerConfiguration server)
```
scottf: This isn't necessary. Both GetChanges/RecordWouldBeChange are specific to comparing consumer configuration when it is being used during a jetstream subscribe.

src/NATS.Client/KeyValue/KeyValueWatchSubscription.cs
```diff
@@ -26,6 +26,12 @@ public class KeyValueWatchSubscription : IDisposable
 
         public KeyValueWatchSubscription(KeyValue kv, IList<string> keyPatterns,
             IKeyValueWatcher watcher, ulong fromRevision, params KeyValueWatchOption[] watchOptions)
+            : this(kv, keyPatterns, watcher, null, fromRevision, watchOptions)
+        {
+        }
```
scottf, line 31: This can't be empty, it has to call the delegate.

src/NATS.Client/KeyValue/KeyValueConsumerConfiguration.cs
MENNAELMESALMY: Yeah, wanted to make it consistent with `ConsumerConfiguration` but that seems unnecessary.

MENNAELMESALMY:
> My initial reaction is that I don't feel good about adding 10 more api. I would be more inclined if you add only 2 more (watch/watch all) with all possible parameters. Also, this has not been approved yet by the client team, I have raised the issue again.

Updated to include only three, I suppose I can make it only for list of keys as well, what do you think?
Thanks for following up with the team, glad to hear their feedback.

src/NATS.Client/KeyValue/KeyValueWatchSubscription.cs
MENNAELMESALMY: Sorry can you clarify more on that? 
I think this `this(kv, keyPatterns, watcher, null, fromRevision, watchOptions) ` just calls the other constructor with the configuration as null.

scottf: Yeah, sorry, I was in Java land a minute ago...

scottf:
@MENNAELMESALMY I have an idea for a solution. Since your main objective is to have the ability to know the consumer name, we could add something to `KeyValueOptions` like `WithConsumerNamePrefix`. We would then use the prefix for any consumers the client creates by itself. By adding it to the options we don't have to change or add any api, but would have to make sure all implementation is properly fixed.
As far as meta data, I understand that this solution does not address that.
Other than that, this idea was voted down by the client team, so I'm not sure what else to offer.

PR #897, opened 2024-05-26
test and fix Nkeys.FromPublicKey

Zetanova:
The currently `Nkeys.FromPublicKey` method produces a `NKeyPair` with an invalid `PublicKey` of 56 bytes

### DIFF
--- a/src/NATS.Client/NKeys.cs
+++ b/src/NATS.Client/NKeys.cs
@@ -11,12 +11,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+
 using System;
 using System.IO;
 using System.Security.Cryptography;
-using System.Text;
 using NATS.Client.NaCl;
-
 namespace NATS.Client
 {
     internal static class Crc16
@@ -115,7 +114,7 @@ namespace NATS.Client
         public byte[] PrivateKeySeed => seed;
 
         public string EncodedSeed => Nkeys.Encode(Nkeys.PrefixFromType(Type), true, seed);
-        
+
         /// <summary>
         /// Wipes clean the internal private keys.
         /// </summary>
@@ -137,7 +136,7 @@ namespace NATS.Client
             return rv;
         }
 
-#region IDisposable Support
+        #region IDisposable Support
         private bool disposedValue = false; // To detect redundant calls
 
         /// <summary>
@@ -164,7 +163,7 @@ namespace NATS.Client
         {
             Dispose(true);
         }
-#endregion
+        #endregion
     }
 
     /// <summary>
@@ -195,7 +194,7 @@ namespace NATS.Client
 
         // PrefixByteUnknown is for unknown prefixes.
         internal const byte PrefixByteUknown = 23 << 3; // Base32-encodes to 'X...'
-        
+
         public enum PrefixType
         {
             User,
@@ -224,7 +223,7 @@ namespace NATS.Client
 
             if (crc != Crc16.Checksum(data))
                 throw new NATSException("Invalid CRC");
- 
+
             return data;
         }
 
@@ -286,7 +285,7 @@ namespace NATS.Client
             // This code commented out b/c string.remove does not touch the original string.
             // There is not way to really wipe the contents of a string.
             // if (src != null && src.Length > 0)
-                // src.Remove(0);
+            // src.Remove(0);
         }
 
         internal static byte[] DecodeSeed(byte[] raw)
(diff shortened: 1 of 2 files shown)


scottf:
Could you provide code or algorithm to reproduce this?

Zetanova:
Use the included unit test. The utility function `Nkeys.FromPublicKey` will never produce a valid NKeyPair instance
[src/Tests/UnitTests/TestNkeys.cs](https://github.com/nats-io/nats.net/blob/0d04b28b1a20da56fec4834d77909c9647e0f427/src/Tests/UnitTests/TestNkeys.cs#L84)

scottf:
@Zetanova This looks good. Could you please add to the main comment a discussion of the primary change, i.e., what was wrong with the old code and why you put in the code that you did.

src/NATS.Client/NKeys.cs
```diff
@@ -137,7 +136,7 @@ public byte[] Sign(byte[] src)
             return rv;
         }
 
-#region IDisposable Support
+        #region IDisposable Support
```
scottf, line 139: Does this indentation matter?

scottf (changes requested):

src/NATS.Client/NKeys.cs
Zetanova: It does not matter. Regions are only for the IDE (collapsible code regions) 
VS2022 did not even show them as changed in the commit
I know it, because I reverted all auto format changes and will try to do it for this one too.

If the project requires a specific format, it needs to be declared with an .editorconfig
reference https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-format

sixlettervariables: Whether you indent region directives is a preference thing; bothers my brain to see things akin to a preprocessor directive indented...but it is legal.

Zetanova: the proof ...
image

Zetanova: I will write to the VS team as a bug, all changes inside a commit should be marked. 

https://developercommunity.visualstudio.com/t/Git-commit-view-window-does-not-show-all/10669950?port=1025&fsid=1a84c04d-2855-4117-a9d3-332895525ca8

src/Tests/UnitTests/TestNkeys.cs
```diff
@@ -79,6 +79,23 @@ public void TestNKEYPublicKeyFromSeed()
...
+            var pubKey1 = userKey1.EncodedPublicKey;
+
+            var userKey2 = Nkeys.FromPublicKey(pubKey1);
+
+            Assert.Equal(userKey1.PublicKey, userKey2.PublicKey);
+
+            var pubKey2 = userKey2.EncodedPublicKey;
+
+            Assert.Equal(pubKey1, pubKey2);
+        }
```
scottf, line 97: Can you please change this test to this. I made it this way so I could have a parallel test in Java and to verify that they were both working the same.

```
[Fact]
public void TestNKEYFromPublicKey()
{
    _TestNKEYFromPublicKey("SUAHBVFYZF3DIEO4UIHIZMJICVLURLBM5JJPK7GSVGP2QUC3NZ323BRE6A", "UCM5BG6AAZSEGREBCLG7PG4GFQNJABSAVIXC6VWS7TDHZFPIYFVYHIDG");
    _TestNKEYFromPublicKey("SAADARCQJ3JA737Z443YNAZBNJNTFP7YNAF4QFUXKTBFBS4KAVK55DGSOQ", "AD2HQTUKOPBUGOPHA6KFRE6ZW5TH43D7P7E56OAQBZQLW2ECMNML6MVA");
    _TestNKEYFromPublicKey("SNAH645525YA4PNXHWWS46VNXXQTYAXOPKGHXYAHXZZ43XTDDG2ZQAX7LY", "NBZCD2OSMSDRVYCAI77HUN6A2WNDWNT2DMVVEW66DHNWCDXVOUWRCCK7");
    _TestNKEYFromPublicKey("SOAF5OP7UPK6XJCMNRYEJRET6YQSOE3FD4I4ERSN6WKHLYUC5AQDCOAFVY", "OA6SJACXYP2QGNLU4QYLJTVRVZPCZEEUNO2UQOVNGXYUPUJJHCVZIZQ2");
    _TestNKEYFromPublicKey("SCAP4LGVURDWVL37AZIM5O47UKANFI6FKBY77HMYF55CKW2XFKLNUBTTFE", "CAO36T42KFA2LMIZ6YHJKPQEJWT5ULYSV633FWBCEJ7MREZPHHC56BSC");
}

private static void _TestNKEYFromPublicKey(string encodedSeed, string encodedPubKey)
{
    NkeyPair fromSeed = Nkeys.FromSeed(encodedSeed);
    NkeyPair fromKey = Nkeys.FromPublicKey(encodedSeed.EncodedPublicKey);

    Assert.Equal(fromSeed.PublicKey, fromKey.PublicKey);
    Assert.Equal(encodedPubKey, fromSeed.EncodedPublicKey);
    Assert.Equal(encodedPubKey, fromKey.EncodedPublicKey);
}
```

scottf (changes requested):

src/Tests/UnitTests/TestNkeys.cs
Zetanova: It's changed now, but with the xunit theory with inline data

scottf (approved):
LGTM

PR #916, opened 2024-08-05
Handling of unobserved tasks after connection opening timeout

TobZip:
Tasks from failed NATS connections appeared when observing and logging unobserved task exception, even though the failed connections were handled, as the result of the connection task wasn't observed by e.g. using `await`  or reading the task's `Exception` property.

This Pull Request changes that by observing the task exception, although the result is then ignored as NATS has already detected the failure as a timeout. The `continueWith` task executed after a connection timeout occurs will now check if there was an exception, preventing the `TaskScheduler.UnobservedTaskException` event from being triggered by an already handled connection failure.

This Pull Request also ensures the TCP client is explicitly closed or disposed even in case of a timeout. This resolves a race condition where the TCP client field typically is set to null before the client is closed or disposed.

A test case was added to ensure an unobserved task exception does not appear when connecting to a NATS server that does not accept connections (because it wasn't started).

That the TCP client now is closed or disposed even in case of a timeout was manually confirmed using breakpoints.

### DIFF
--- a/src/NATS.Client/Connection.cs
+++ b/src/NATS.Client/Connection.cs
@@ -394,7 +394,16 @@ namespace NATS.Client
                         if (!task.Wait(TimeSpan.FromMilliseconds(options.Timeout)))
                         {
                             nce = new NATSConnectionException("timeout");
-                            task.ContinueWith(t => close(client));
+
+                            // The client field may be sets to null before ContinueWith executes and calls close.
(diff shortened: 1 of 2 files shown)


sixlettervariables (approved):

scottf:
@TobZip Thank you for this contribution

PR #918, opened 2024-08-08
Netstandard 2 build changes

scottf:
We are removing support for `netstandard1.6` and `net46` targets and will only support the `netstandard2.0` target. This change is intended to simplify maintenance. As a result, some platforms will no longer be supported. For more information, please refer to the compatibility table provided by Microsoft: [Select .NET Standard version](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0#select-net-standard-version).


mtmk (approved):
LGTM

PR #919, opened 2024-09-04
Add support for clearing old certificates during reconnect

TobZip:
The `Options` object does not provide a way to remove once added certificates.
This leads to that in case of certificate renewals the connection needs to be closed and recreated, including recreating all subscriptions. Just adding the new certificate and trying to reconnect with the old no longer relevant certificate and the new certificate in the certificate collection will make the connection fail on the TLS handshake.

Without a way to clear the certificates, reflection is required to get access to the internal `certificates` field which can be problematic.

By enabling library user to clear the certificates field through `ClearCertificates()` and adding the new certificate, the connection can use the correct certificate during the next connection retry (if retrial is configured).
This can be done for example in a user provided `ReconnectDelayHandler` method.

Other changes: 
- fixed some typos
- added missing usage of `nameof()`

### DIFF
--- a/src/NATS.Client/Options.cs
+++ b/src/NATS.Client/Options.cs
@@ -792,6 +792,18 @@ namespace NATS.Client
         }
 
         /// <summary>
+        /// Clears all X.509 certificates added through <see cref="AddCertificate(string)"/>.
+        /// </summary>
+        /// <remarks>
+        /// Can be used to renew the client certificate without recreating the <see cref="Connection"/>.
+        /// </remarks>
+        public void ClearCertificates()
+        {
+	        certificates?.Clear();
+        }
+
+
+        /// <summary>
         /// Overrides the default NATS RemoteCertificationValidationCallback.
         /// </summary>
         /// <remarks>
--- a/src/NATS.Client/Options.cs
+++ b/src/NATS.Client/Options.cs
@@ -21,7 +21,7 @@ using static NATS.Client.Defaults;
 namespace NATS.Client
 {
     /// <summary>
-    /// This class is used to setup all NATs client options.
+    /// This class is used to set up all NATs client options.
     /// </summary>
     public sealed class Options
     {
@@ -188,7 +188,7 @@ namespace NATS.Client
         /// </summary>
         /// <remarks>
         /// A server in lame duck mode will gradually disconnect all its connections
-        /// before shuting down. This is often used in deployments when upgrading
+        /// before shutting down. This is often used in deployments when upgrading
         /// NATS Servers.
         /// </remarks>
         public EventHandler<ConnEventArgs> LameDuckModeEventHandler;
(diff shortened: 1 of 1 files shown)


src/NATS.Client/Options.cs
```diff
@@ -287,8 +287,8 @@ public void SetUserCredentialsFromString(string credentialsText)
         /// Sets user credentials from text instead of a file using the NATS 2.0 security scheme.
         /// </summary>
         /// <param name="userJwtText">The text containing the "-----BEGIN NATS USER JWT-----" block</param>
-        /// <param name="nkeySeedText">The text containing the "-----BEGIN USER NKEY SEED-----" block or the seed begining with "SU".
-        /// May be the same as the jwt string if they are chained.</param>
+        /// <param name="nkeySeedText">The text containing the "-----BEGIN USER NKEY SEED-----" block or the seed beginning with "SU".
+        /// Maybe the same as the jwt string if they are chained.</param>
```
scottf, line 291: "May be" was correct here. How about, "Both string can contain both the credentials and jwt block since they will be parsed for the required data."

src/NATS.Client/Options.cs
```diff
@@ -784,13 +785,25 @@ public void AddCertificate(string fileName)
...
+
+        /// <summary>
+        /// Clears all X.509 certificates added through <see cref="AddCertificate(string)"/>.
+        /// </summary>
+        /// <remarks>
+        /// Can be used to renew the client certificate without recreating the <see cref="Connection"/>.
+        /// </remarks>
+        public void ClearCertificates()
+        {
+	        certificates?.Clear();
```
scottf, line 803: Options is "intended" to be immutable. Just make another options object.

scottf (changes requested):
So I'm not going to accept this because it's not necessary. Just make another Options. I do like the other changes, the comment fixes and the nameof for instance, so you are welcome to resubmit with only those changes.

src/NATS.Client/Options.cs
TobZip: Ok, but how do I use this new options object when I have an existing connection? The only way I see is to create a new connection through the connection factory and pass as parameter in `CreateConnection`, or?
I want to avoid recreating the connection object based on the reasons I wrote in the PR description.

scottf: Ok, I'm convinced, accepting the PR. Thank you for the contribution and the feedback.

scottf (approved):
LGTM

src/NATS.Client/Options.cs
TobZip: Great, thanks for your fast answer!

src/NATS.Client/Options.cs
TobZip: I did not have time to make this change according to your suggestion before this PR was merged. 
Do you want me to make a new PR to fix this?

scottf: sorry for the slow response, yeah, please fix when you get a chance. I'll make a release as soon as you do

TobZip: same here, sorry for the even slower response. Will make a new PR this week.

TobZip: PR is now created: #925

PR #921, opened 2024-10-07
Concurrent Dictionary for thread safety

scottf:
Addressing https://github.com/nats-io/nats.net.v1/issues/917

### DIFF
--- a/src/NATS.Client/Connection.cs
+++ b/src/NATS.Client/Connection.cs
@@ -2930,8 +2930,7 @@ namespace NATS.Client
         {
             lock (mu)
             {
-                InFlightRequest ignored;
-                waitingRequests.TryRemove(requestId, out ignored);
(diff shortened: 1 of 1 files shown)


src/NATS.Client/Connection.cs
```diff
@@ -4481,15 +4489,15 @@ internal void removeSubSafe(Subscription s)
         // caller must lock
         internal virtual void removeSub(Subscription s)
         {
-            subs.Remove(s.sid);
+            Subscription ignore;
+            subs.TryRemove(s.sid, out ignore);
```
mtmk: you could use `subs.TryRemove(s.sid, out _);` if possible

mtmk (approved):
LGTM

src/NATS.Client/Connection.cs
scottf: Changed

PR #925, opened 2024-11-18
Improve Options documentation

TobZip:
Related to comments on PR #919 which were missed before merge to main.


scottf (changes requested):
LGTM

PR #934, opened 2025-05-02
Update ApiStats, Account Tier and AccountStatistics

scottf:


src/NATS.Client/JetStream/AccountStatistics.cs
```diff
@@ -21,34 +22,44 @@ namespace NATS.Client.JetStream
...
     {
         // rollup contains the memory, storage, streams, consumers and limits from the top level
-        private AccountTier _rollup;
+        private AccountTier _rollupTier;
         
         /// <summary>
         /// Gets the amount of memory storage used by the JetStream deployment.
         /// </summary>
-        public long Memory => _rollup.Memory;
+        public ulong MemoryBytes => _rollupTier.MemoryBytes;
```
scottf, line 30: replaces obsolete Memory that was long which does not handle unit64

mtmk (approved):
LGTM

src/NATS.Client/JetStream/AccountStatistics.cs
```diff
@@ -21,34 +22,44 @@ namespace NATS.Client.JetStream
...
         
         /// <summary>
         /// Gets the amount of file storage used by the JetStream deployment.
         /// </summary>
-        public long Storage => _rollup.Storage;
+        public ulong StorageBytes => _rollupTier.StorageBytes;
```
scottf, line 35: replaces obsolete that was long which does not handle unit64

PR #937, opened 2025-05-08
NuGet note to deprecate the package

mtmk:
We had a few reports people still landing on this package by just using NuGet interfaces on Visual Studio. This note should help guide some of those new developers.

**note** if we don't want to say 'deprecated' we can say 'we suggest' or something to that effect if you like.

**note2** also the current text is not accurate since now v2 supports netstandard as well, so 'net6 and above' is not right


scottf (approved):
LGTM

PR #938, opened 2025-05-15
Fix header value validation

scottf:
Like rfc822 section 3.1.2 (quoted in ADR 4)
The field-body may be composed of any US-ASCII characters, except CR or LF.


mtmk (approved):
LGTM

mtmk (approved):
LGTM

PR #942, opened 2025-06-05
Handle Leadership Change and Server Shutdown statuses

scottf:
Added handling of Leadership Change. Workflow matches Java.


mtmk (approved):
LGTM

PR #945, opened 2025-08-27
Support ordered consumer name prefixing

scottf:
1. Ordered consumers use the consumer configuration name as a prefix. (They can't use the name in case there is an communication error / sequence ordering error and have to make a new consumer). Prefix is designed to help find consumer names in logging. The old way, the name was always random.

2. Key Value and Object Store watchers use ordered consumers under the covers. To have a consumer name prefix you can implement the new extended interfaces
```
public interface IKeyValueWatcherPrefixable : IKeyValueWatcher
public interface IObjectStoreWatcherPrefixable : IObjectStoreWatcher
```


mtmk (approved):
LGTM

PR #948, opened 2025-09-11
ObjectMeta metadata

scottf:
The client has ObjectMeta Headers, which is a dictionary of string (key) to list of strings (value), but does not have the additional, almost identical field of metadata, which is a dictionary of string (key) to string (value).


mtmk (approved):
LGTM

